Issue #3, opened 2018-06-12
add template to repo


Mathadon:

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/4.

PR #4, opened 2018-06-13
added initial version of MPC template

Mathadon:


dhblum:
Just to be clear, this is intended to be a template for an emulator model, not an MPC controller, correct?  If so, I suggest changing the name of the model to EmulatorTemplate?

Mathadon:
good point, fixed!

dhblum:
Closes https://github.com/ibpsa/project1-boptest/issues/3.

PR #7, opened 2018-08-06
Revert "Add the overwritten block to the test case"

dhblum:
@SenHuang19 I think it is better not to commit the whole IBPSA library to this repository.  Otherwise, it will be difficult to maintain as the IBPSA Library continues to be developed.  In addition, I think once we merge changes to the master, it would be better to not have the folder "From_senhuang", and instead a more integrated file directory.  The differences between what we had before and additions will be tracked in the commit history.

Therefore, I am proposing to revert this merge with the master.


SenHuang19:
@dhblum I found there is one issue in using the overwritten block with JModelica. 

In the overwritten block, we used the sample function to trigger the c function to send/receive data from external sources, which works quite well in dymola environment under both Linux and Window OS . 

However, with Jmodelica, the c function is triggered multiple times at the same simulation time. I tried the following methods to resolve the issue:
- converting the sample function into a equivalent if-statement;
- converting the Modelica model into a whole model;
- reducing the number of the ovwritten blocks to 1;

However, none of the methods resolve the problem.
One thing I noticed though is that the "time" array in the simulation results from JModelica does have the duplicated values. I am not sure if that is related to the issue above.

Issue #9, opened 2018-09-11
Add szvav


dhblum:
This issue is to add a test case that contains a more realistic building model.  It is based on the single zone VAV system model in the Buildings Library (Buildings.Air.Systems.SingleZone.VAV.ChillerDXHeatingEconomizer).   The prototype code for building and deploying is based on the initial test case.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/10.

PR #10, opened 2018-09-15
Issue9 szvav

dhblum:
This adds a single zone vav test case and documentation.  It also restructures  the directories of the repository to contain two test cases, 1 and 2.  Testcase1 is the original simple RC, while Testcase2 is the new Single Zone VAV.


dhblum:
Closes https://github.com/ibpsa/project1-boptest/issues/9.

Issue #11, opened 2018-09-15
Abstract test cases


dhblum:
Right now, each test case has duplicate code for Docker image build and deployment. This issue is to abstract the test cases in a way that reduces code duplication.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/12.

PR #12, opened 2018-09-20
Issue11 abstract test cases

dhblum:


dhblum:
Closes https://github.com/ibpsa/project1-boptest/issues/11.

Issue #23, opened 2019-01-30
Parser fails with final parameter


dhblum:
This issue is to fix the parser failing when the parameters in the IBPSA signal exchange blocks have the final keyword.  Introduced by https://github.com/ibpsa/modelica-ibpsa/commit/06a73634d1899843b2b35a92b89b7a350597c57a.

It is likely that the search for the parameters by causality by ``fmu.get_model_variables(causality = 0)`` needs to be changed.

dhblum:
@mwetter FYI.

dhblum:
@mwetter For the get_model_variables(causality = <#>) function in pyfmi, the below is given for defining which type of variable is returned.  Any idea what "Local" refers to?
```
        The causality of the variables (Parameter==0,
        Calculated Parameter==1, Input==2, Output==3, Local==4,
        Independent==5, Unknown==6).
        Default: None (i.e all).
```

mwetter:
@dhblum From https://fmi-standard.org/docs/2.0.1-develop/#_definition_of_model_variables_modelvariables
> - "local": Local variable that is calculated from other variables or is a continuous-time state (see section 2.2.8). It is not allowed to use the variable value in another model or slave.

dhblum:
Ah ok, thanks for the reference.  

It looks like if in a Modelica model compiled by JModelica you use ``parameter par_name``, ``par_name`` is returned with causality = 0.  

If instead you have ``final parameter par_name``, ``par_name`` is returned with causality = 4.

mwetter:
You could use something like
```modelica
model Test
  parameter Real par = 2;
  final parameter Real par_final = 2;
  Real x1(start=1);
equation
  der(x1) = - par*par_final*x1;

 annotation (experiment(Tolerance=1e-6, StopTime=1.0));
end Test;
```
and then
```
In [7]: mod.get_variable_variability('par')
Out[7]: 1

In [8]: mod.get_variable_variability('par_final')
Out[8]: 0

In [9]: mod.get_variable_variability('x1')
Out[9]: 4
```

```

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/27.

PR #24, opened 2019-02-08
added causality=4 to make fix

JavierArroyoBastida:
Addresses https://github.com/ibpsa/project1-boptest/issues/23


dhblum:
Thanks @JavierArroyoBastida.  It actually looks like ``vars`` is an inbuilt python function, therefore, I suggest using a different variable name.  Also, please update this branch with the recent changes to the origin/master.  You should be able to do this by pulling the latest origin/master, and merging it to JavierArroyoBastida/issue23_finalParFail.

Issue #25, opened 2019-02-09
Create a contributors list


dhblum:
This issue is to create a contributors list to credit those individuals and institutions who have made contributions to the repository.

PR #26, opened 2019-02-09
Create contributors.md

dhblum:
Address issue https://github.com/ibpsa/project1-boptest/issues/25.


dhblum:
@mwetter , can you please review?

mwetter:
@dhblum It would be good to also include the PNNL and NREL team, and Lieve as they contributed too, not necessarily with software but with design.

dhblum:
Yes of course, I've committed an updated list.

PR #27, opened 2019-02-12
changed vars name

JavierArroyoBastida:


dhblum:
Thanks @JavierArroyoBastida.  I'm sorry to request another change, but actually, if we use ``fmu.get_model_variables(variability = 0)`` and ``fmu.get_model_variables(variability = 1)`` as Michael shows here: https://github.com/ibpsa/project1-boptest/issues/23, we can get the same result, with less variables to sort through.  Using the ``variability`` keyword instead of ``causality`` will give us only parameters and not parameters and all state variables.

JavierArroyoBastida:
@dhblum done!

Issue #22, opened 2019-01-30
Testing


dhblum:
This issue is to implement unittesting.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/28.

PR #28, opened 2019-02-14
Issue22 testing

dhblum:
Addresses https://github.com/ibpsa/project1-boptest/issues/22 by:

- Allowing deployment of testcase# in detached mode
- Adding tests in python using unittest package for each testcase# and parser
- Adding  reference results and other unittest utilities
- Add report generator that collects results from all tests and displays to user.


dhblum:
Test reporter also now returns exit code to make file.

Issue #48, opened 2019-04-02
Empty input data to FMU


dhblum:
If the FMU has inputs but at least one is not written by an external controller, which is ok in the case of overwrite signals or input signals with start values, the simulation by ``testcase.py`` fails due to the lines
```
                for key in u.keys():
                    if key != 'time' and u[key]:
                        value = float(u[key])
                        u_list.append(key)
                        u_trajectory = np.vstack((u_trajectory, value))
                input_object = (u_list, np.transpose(u_trajectory))
```
since the code assumes there is a key that is to be overwritten, creating an ``input_object`` with a time trajectory and no other data, which is unacceptable by pyfmi.

The fix is adding a check as to whether an input object is needed.  If not, make ``None``.

dhblum:
Applied fix in https://github.com/ibpsa/project1-boptest/tree/issue48_emptyInputData.

Need to implement testing.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/53.

PR #53, opened 2019-04-08
Issue48 empty input data

dhblum:
Addresses https://github.com/ibpsa/project1-boptest/issues/48.


dhblum:
Closes https://github.com/ibpsa/project1-boptest/issues/48.

PR #54, opened 2019-04-09
modify the julia code to remove the sleep function

SenHuang19:


dhblum:
Fixed instead by https://github.com/ibpsa/project1-boptest/pull/56.

Issue #17, opened 2018-12-05
Example Julia controller


dhblum:
This issue is to implement an example controller that performs the same tasks as ``examples/twoday-p.py`` and ``examples/szvav-sup.py`` in testing with testcase1 and testcase2.

To Do:
- [x] Initial implementation
- [x] Add unit test.
- [x] Arrange directory structure in examples to make clear distinction between Julia examples and Python examples.
- [x] Fix sleep issue
- [x] Replace docker compose with make file.

SenHuang19:
1. a new branch is created for this issue;
2. initial implementation is pushed;
3. A docker compose structure is added, let me know if we want to use the docker compose to replace the current make files

dhblum:
The following is still needed: 
[] Add unit test.
[] Arrange directory structure in examples to make clear distinction between Julia examples and Python examples.

dhblum:
Sleep issue addressed in https://github.com/ibpsa/project1-boptest/pull/56.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/69.

PR #58, opened 2019-04-22
Issue17 example julia controller

dhblum:
Addresses https://github.com/ibpsa/project1-boptest/issues/17 for making changes to directory structure, adding tests, and a few other documentation edits.


dhblum:
@SenHuang19, I've suggested some changes to the directory structure to make it clear which example controllers are for which language (namely separating into ``python`` and ``julia`` folders).  I also added unittests for the julia controllers and edited some of the documentation.  Can you review the changes, determine if they are ok with you, and suggest any further issues that you think need to be addressed?  I think this is in pretty good shape though and don't have any.

SenHuang19:
1. for the unit test, I suggest to add "docker-compose down" to the makefile to remove the julia image when the test completes;
2. for the julia docker, I suggest we use the port forwarding rather than using the host network to avoid any future issues;
3. if we aren't going to use docker compose to connect controllers and boptest, I suggest to create a makefile for the julia docker as well rather than using docker compose.

dhblum:
Thanks Sen for the review and suggestions.

With regards to 3., my argument against using Docker compose to connect all controllers with boptest is the additional layer of complexity this adds to users because they would also have to specify a Dockerfile for their controller.  Seeing examples of how they may connect controllers to boptest simply through http requests is a lower barrier to entry, as we do with our Python examples.  I think this is ok, as Python is a commonly used language now.  Julia is not as widely-used as Python, so a Docker container is a good way to show a different controller, without burdening a would-be user with downloading Julia.  Your setup also shows for more advanced users how a Docker container with a controller can be connected to boptest if wanted.  Therefore, I think the Python and Julia examples can be used to show both basic and advanced use of boptest.  What do you think?

Issue #55, opened 2019-04-10
Data Generator Module


dhblum:
This issue addresses the development of a data generator, which is used to coordinate the use of test case data for the Forecast and KPI Calculation modules, as well as export of an FMU from a test case directory.

To Do:
- [x] Getting variable data for Forecast and KPI Calculation from multiple CSVs in the FMU
- [x] Explore addition to parser to integrate FMU export from test case.
- [x] Finish definition of variable names and add check that KPI json output variables have corresponding values in data csv (for instance for Forecaster to retrieve)
- [ ] Create 2nd year weather data for MPC model identification

JavierArroyoBastida:
@dhblum, a few notes from our discussions this week to keep them in the history. @mwetter could you please review them?

- The `generate_data` method of the data generator should be called from the `export_fmu` method of the parser since this method needs to be executed only once in the creation of the test case. 
- The `generate_data` method will inspect the variables of all files in the testcase#/models/Resources folder to look for test case data. Every file name is possible but a convention is stablished for the variables to be integrated in the test case data and all files should contain a time column. 
- The data will be stored within the resources folder of the wrapped.fmu in different CSVs with different sampling time periods.
- The sampling period to be used is the one provided by the original data file in the testcase#/models/Resources folder. 
- A forward fill resampling will be applied for all variables except for the weather data.
- The weather data will be provided in a sampling of one minute since it requires spline interpolation that is computed in the weatherTMY3 reader. @dhblum and I concluded that re-implementing the spline interpolation in the forecaster may be an unnecessary code duplication.
- The generation of data for system identification should allow the user to use his/her own control input signals but under different boundary conditions than the standard boundary conditions of the test case. For that the following is required:
  1. The generation of a second weather file from the first one. To be done with the code proposed by Krzysztof.
  2. Enabling wrapped.fmu to dynamically pick the desired weather file (either weather standard or weather system id) when calling `advance`. This has proven particularly challenging because the absolute path is defined in the model compilation. For the data generator we hard-code a change in the filNam parameter of the ReaderTMY3 model to point to the file before compiling ReaderTMY3.fmu . However, here we cannot do the same because wrapped.fmu is already compiled. Any idea how to solve this @mwetter?

dhblum:
2nd year weather data for MPC model identification will be addressed in its own issue.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/61.

PR #61, opened 2019-04-27
Issue55 data generator module

JavierArroyoBastida:
This is for https://github.com/ibpsa/project1-boptest/issues/55


dhblum:
Thanks for this @JavierArroyoBastida .  A few comments:

The test ``test_load_data_and_kpisjson`` does not check that the content of the kpis.json, only that it exists.  

Also, in the ``Data_Manager`` class, ``get_data()`` first checks that data is loaded, and if it is not loaded, then loads the data.  I think instead, data should get loaded on initialization of the class, which is a more intuitive time to load the data for future retrieval by other modules that need it.  Otherwise, if for some reason it takes time to load the data, a user or future developer may wonder why the first call of ``get_data ()`` takes longer than subsequent calls.

Therefore, I suggest the following to do:
- [] Check content of kpis.json in ``test_load_data_and_kpisjson``.
- [] Move the loading of data to the constructor of the ``Data_Manager`` class.

JavierArroyoBastida:
@dhblum with the last commit I'm checking the content of kpis.json in `test_load_data_and_kpisjson`.
For the loading of data I just realized it is done in the constructor of the testcase:
https://github.com/JavierArroyoBastida/project1-boptest/blob/a0855adb7cb1525808c7d0a571be925a4a48296d/testcase.py#L38
So the check in `get_data()` is simply not needed, I'd just remove it.

dhblum:
Thanks @JavierArroyoBastida .  Please, just a few more housekeeping items:
- [ ] Merge the latest ibpsa master.
- [ ] Add a readme to ``/data`` describing the purpose of the modules within.
- [ ] Update the ``/testing/README.md`` to include the data generator tests. 

I am going to make some additional edits to clarify some documentation I've noticed in the code.

JavierArroyoBastida:
@dhblum I have included the items you were proposing. I believe it is ready to merge once you add your edits in the documentation. Please let me know if you find something else.

Issue #36, opened 2019-02-25
Continuous integration testing


dhblum:
This issue is to add continuous integration testing with [Travis](https://travis-ci.org/).

dhblum:
@mwetter I've pushed an initial implementation of a travis yml to a branch (https://github.com/ibpsa/project1-boptest/tree/issue36_ciTravis, see https://github.com/ibpsa/project1-boptest/commit/74c9782c6696c0f7ad1cc4813426dad0cb6509fa).  How do you recommend I initiate travis and test the ci?

mwetter:
@dhblum : I enabled CI testing, please see https://travis-ci.org/ibpsa/project1-boptest/branches

dhblum:
Great, thank you.

dhblum:
https://github.com/ibpsa/project1-boptest/tree/issue36_ciTravis_b is passing.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/79.

dhblum:
Add build icon.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/81.

PR #79, opened 2019-06-13
Issue36 ci travis b

dhblum:
Addresses https://github.com/ibpsa/project1-boptest/issues/36.


dhblum:
Also addresses https://github.com/ibpsa/project1-boptest/issues/80.

Issue #87, opened 2019-07-16
Update example test case data


dhblum:
This issue is to update the boundary condition data in testcase1 and testcase2 to match model implementations.

PR #88, opened 2019-07-16
Issue87 update test case data

dhblum:
Addresses https://github.com/ibpsa/project1-boptest/issues/87.


dhblum:
@JavierArroyoBastida Can you please review?

dhblum:
@JavierArroyoBastida ``zon.intLoad`` gives the schedule of 0.1 during unoccupied hours and 1.0 during occupied hours.  ``zon.qRadGai_flow``, ``zon.qConGai_flow``, ``zon.qLatGai_flow`` give the nominal values of 288, 192, 96 respectively in W (note that the area of the room is 48 m2 and the detailed room model in ``Buildings`` requires the internal gain inputs to be W/m2).  So the total internal gain at schedule value of 1.0 is 576.  Also, for occupancy, one person doing very light to moderate office work in an office or apartment is 45-60 W latent, so my assumption is that when occupied, there are 2 people in the zone.

JavierArroyoBastida:
@dhblum that clarifies everything. Thanks for the detailed response. I'd go ahead with the merge.

PR #95, opened 2019-08-05
1. reconfigure the customized KPI input; 2. add readme

SenHuang19:


dhblum:
@SenHuang19 I think the naming of the files could be changed to make more clear the division between what code is generic/required for all cases, and what code is only serving as an example implementation of custom KPI calculation.  I suggest the following.  What do you think?

- rename ``kpicalculation.py`` to ``custom_kpi_calculator.py``
- rename ``project1-boptest/examples/python/kpi`` to ``project1-boptest/examples/python/custom_kpi``
- rename ``kpi_example.py`` to ``custom_kpi_szvav_sup.py``
- rename ``kpi.config`` to ``custom_kpi_szvav_sup.config``

SenHuang19:
Added the unit test files

Issue #100, opened 2019-08-24
Split internal gain data into rad, con, and lat


dhblum:
Internal load data should be represented for each zone in the boundary condition data as split into radiant, convective, and latent portions.  This will make it easier to represent the internal heat gains needed for MPC controllers, especially since the ratio of rad/con/lat can change over time based on occupant number and types of equipment that are in operation.

PR #63, opened 2019-04-27
Issue35 kpi calculation module

JavierArroyoBastida:
This is for https://github.com/ibpsa/project1-boptest/issues/35

PR #101, opened 2019-08-27
Issue100 int load split

dhblum:
This is for https://github.com/ibpsa/project1-boptest/issues/100.


JavierArroyoBastida:
@dhblum this is fine by me. https://github.com/ibpsa/project1-boptest/pull/63 will have to be updated accordingly if this one is merged first.

Issue #115, opened 2019-10-21
Deal with models without Overwrite blocks


SenHuang19:
This issue is to deal with models without Overwrite blocks by modifying the parse function.

SenHuang19:
If models have no Overwrite blocks, the following error occur when parsing the model

> Unexpected token: "end wrapped"

A quick solution is to change line 157-160 of the parse

>             if i == len(instances['Overwrite'])-1:
>                 f.write(') "Original model with overwrites";\n')
>             else:
>                 f.write(',\n')

into

>             if i != len(instances['Overwrite'])-1:
>                 f.write(',\n')
>       f.write(') "Original model with overwrites";\n')

Not sure if other components have to be updated as well? @dhblum

dhblum:
@SenHuang19 What do you think about just exporting the original model fmu, instead of writing what is essentially an empty wrapped.mo?  I've implemented suggested code in https://github.com/ibpsa/project1-boptest/pull/118.

SenHuang19:
@dhblum The wrapped.mo may not be empty as there may be the "Modelica.Blocks.Interfaces.RealOutput"s for "out read". If we just export the original model fmu , the generated fmu is not capable with the BOPTEST APIs, right?

dhblum:
When loaded into BOPTEST, any Modelica.Blocks.Interfaces.RealOutput or Modelica.Blocks.Interfaces.RealInput in the generated FMU will still be recognized by BOPTEST, and actually available for reading/writing.  However, the default value of inputs will be 0 (unless otherwise specified with a start value in the model), they will not be associated with an activate/deactivate signal, and the outputs will not be linked to any KPI calculations.

SenHuang19:
@dhblum thanks for clarifying this. It is my view that it is necessary to make sure the generated FMU can be supported by all the APIs including KPIs.

dhblum:
The only issue I see here with the BOPTEST API is that using the ``parser.py`` to compile a Modelica model with no signal exchange blocks into an FMU will generate an empty ``kpis.json`` file.  If this empty ``kpis.json`` is included in the testcase FMU, then the ``/kpi`` API call fails because there are no outputs specified to calculate the core KPIs.  That is why I said if no signal exchange blocks are used AND the user wants to still calculate core KPIs using FMU outputs defined in their own way, the content of the ``kpis.json`` needs to be generated a different way.

[...]

PR #118, opened 2019-10-31
Issue115 no signal exchange block

dhblum:
For https://github.com/ibpsa/project1-boptest/issues/115.


dhblum:
@SenHuang19 I added support and tests for models with only read blocks.  So this PR now addresses models with no signal exchange blocks, and only read blocks.  Can you review before merging?

SenHuang19:
@dhblum thanks for the information. The fix looks good to me and I just have one minor comments: 

 - line 170 of parser.py:
       `                f.write(') "Original model with overwrites";\n')`

    can be changed into:
       `                f.write(') "Original model without overwrites";\n')`

dhblum:
Thanks @SenHuang19, I've made the change you suggest and merged the latest master.  I will merge after checks pass.

Issue #117, opened 2019-10-25
Add customized starting time for simulation


SenHuang19:
This issue adds functions to allow users to specify the starting time when resetting the simulation

dhblum:
This issue will change the ``reset`` api call to ``initialize`` to be more appropriate for the functionality.

PR #126, opened 2019-11-18
modify the reset API to take inputs from users regarding the starting time

SenHuang19:
this is related to #117

### DIFF
--- a/restapi.py
+++ b/restapi.py
@@ -36,6 +36,9 @@ parser_step.add_argument('step')
 parser_advance = reqparse.RequestParser()
 for key in case.u.keys():
     parser_advance.add_argument(key)
+# ``reset`` interface    
+parser_reset= reqparse.RequestParser()
+parser_reset.add_argument('reset_time')
 # -----------------------
 
 # DEFINE REST REQUESTS
@@ -55,7 +58,9 @@ class Reset(Resource):
     
     def put(self):
         '''PUT request to reset the test.'''
-        case.reset()
+        args=parser_reset.parse_args()
+        reset_time=args['reset_time']
+        case.reset(reset_time)
         return 'Testcase reset.'
 
         
--- a/testcase.py
+++ b/testcase.py
@@ -120,12 +120,29 @@ class TestCase(object):
         
         return self.y
 
-    def reset(self):
+    def reset(self,reset_time):
         '''Reset the test.
         
-        '''
+        Parameters
+        ----------
+        reset_time : float
+            Simulation start time in seconds.
+            
+        Returns
+        -------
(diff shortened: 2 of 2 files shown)


dhblum:
@SenHuang19 Your suggested method implements a warm-up period by simulating from 0 to reset_time, to eliminate the effect of initial conditions, correct?  Should the behavior instead be similar to setting a new simulation start time in pyfmi or dymola, where this function just resets self.start_time, instead of first simulating the model from 0 to the reset time?  My arguments for this are that then the behavior is the same as expected in these other tools, the behavior is the same whether the simulation starts from 0, 1 day, or 300 days in terms of treatment of initial conditions (or lack there-of for now), and it does not require the time and data of simulation of 0 to 300 days if the user wants to start the simulation at 300 days.  What do you think?

SenHuang19:
> @SenHuang19 Your suggested method implements a warm-up period by simulating from 0 to reset_time, to eliminate the effect of initial conditions, correct? Should the behavior instead be similar to setting a new simulation start time in pyfmi or dymola, where this function just resets self.start_time, instead of first simulating the model from 0 to the reset time? My arguments for this are that then the behavior is the same as expected in these other tools, the behavior is the same whether the simulation starts from 0, 1 day, or 300 days in terms of treatment of initial conditions (or lack there-of for now), and it does not require the time and data of simulation of 0 to 300 days if the user wants to start the simulation at 300 days. What do you think?

1. The reason to start the simulation from 0 is to make sure the simulation result is kind of consistent with similar initial states, regardless of when to start the simulation;
2. I acknowledge that it may take significant time to do this kind of reset 
3. Alternatively, may I suggest to create dynamic warming up periods. Specificially, if the user wants to reset the simulation to be starting from June 30, the warning period will be starting from June 28.

dhblum:
This is an interesting point.  

My concern is us making assumptions about the necessary length of warming periods for all models and test scenarios.  My view is that this should be left to the user and/or to the specifiers of test scenarios, to be based on their specific application(s).  Then, the value assigned to the reset function can be specified based on their warming requirements.  All we do here is provide the functionality to specify it.

Or, are you already suggesting to allow a user to specify both a start time and a warming period length?

dhblum:
Just to document, I think the design we landed on was that the API call would take start time and warm up period as arguments.

dhblum:
@SenHuang19 I've created a new development branch for this at https://github.com/ibpsa/project1-boptest/tree/issue117_resetTime.  Please make a PR to this branch.

dhblum:
Closing in favor of https://github.com/ibpsa/project1-boptest/pull/173.

Issue #97, opened 2019-08-20
Move design and requirements document from google doc


dhblum:
This issue is to move the design and requirements document from a google doc to rst-based documentation in this repository.

dhblum:
@yanchenpnnl To work on update for KPIs and associated Appendix.

dhblum:
@yanchenpnnl Development is on branch https://github.com/ibpsa/project1-boptest/tree/issue97_docDesign.

dhblum:
The documentation will be moved as written in google docs today (12/24/2019).  Additional necessary updates will be made with future issues.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/140.

SenHuang19:
I suggest we to create a github wiki for showing this document

dhblum:
Can we easily manage the wiki using rst files?

PR #139, opened 2019-12-24
Issue97 doc design

yanchenpnnl:
Added the core KPI section and Appendix section for additional discussion about other KPIs


SenHuang19:
just one suggestion: should we create a wiki rather than asking users to generate the pdf or html file locally?

dhblum:
@yanchenpnnl I made a few edits, see the commits.  Mainly addressing compilation warnings/issues and layout.  I think it is ready to merge to the main development branch.  Additional suggested edits can be made with new pull requests to the development branch.

@SenHuang19 If you mean a sort of read-the-docs from the rst files, I think that would be good.  I have not set that up before though.  Please continue discussion in https://github.com/ibpsa/project1-boptest/issues/97.

Issue #107, opened 2019-09-10
Add CO2 KPI to KPI Calculation Module


dhblum:
This issue is to add the IAQ KPI to the KPI calculation module by introducing CO2 ppm-hour calculations.  This also requires addition of a threshold in test case data as well as CO2 species concentrations to test case emulation models.

PR #133, opened 2019-12-06
Issue121 specify zone for data

JavierArroyoBastida:
For https://github.com/ibpsa/project1-boptest/issues/121
@dhblum could you please review? Before sending a pull request to the master I would:
1. Merge https://github.com/ibpsa/modelica-ibpsa/tree/issue1257_zoneDesignation to master.
2. Revert commit https://github.com/JavierArroyoBastida/project1-boptest/commit/c1b9dd52764e9da15a747ef29e8e05daf6a4c885 that is checking out issue1257_zoneDesignation of IBPSA.
3. Develop a unit tests for checking functionality when multi-zone specifications are used.

PR #152, opened 2020-03-05
Issue107 iaq kpi

dhblum:
For https://github.com/ibpsa/project1-boptest/issues/107.

- CO2 tracking and measurements added to testcase1 and testcase2.
- Upper limit for CO2 added to test case data and design requirements documentation (header called ``UpperCO2``).  Value based on ASHRAE Guideline 36 Table 3.1.1.3.
- IAQ KPI calculation based on CO2 measurements added to KPI calculator (result called ``idis_tot``).
- Edits testcase1 and testcase2 to have non-zero iaq kpi calculations.
- Unittest results updated.


dhblum:
@JavierArroyoBastida Can you review this when you get chance?

JavierArroyoBastida:
@dhblum this is a very nice progress, thanks for working on this! My only two concerns:
1. I see KPIs for IAQ of actual test cases (testcase1 and testcase2) are both 0 which may not trigger any warning if something in the code affects IAQ calculation. 
2. Do we want to allow different CO2 thresholds for different zones? if yes that's to be worked out in https://github.com/ibpsa/project1-boptest/pull/133. In any case I suggest to merge this pull request already to the master and I'll update https://github.com/ibpsa/project1-boptest/pull/133 with these latest changes. We could even make use of the new multi-zone testcase  that needs to be developed on the issue121 branch as a test case with some ppm*h to address point 1.

dhblum:
Thanks @JavierArroyoBastida for the review.  Yes, I think that it would be good to allow different thresholds for different zones, since different zone types may have different limits, or test case developers may develop to different thresholds.  This is similar to temperature setpoints.  Yes, it can be addressed in #133.

To your first point, before merging, I propose adding an arbitrary CO2 concentration to testcase1 that violates the limits for some amount of time in order to test the calculation of KPI.  I propose a sinusoidal CO2 concentration varying between 500 and 1000 diurnally.  Since the limit is 894, it will violate for some times.  I will make a new push soon for this.

dhblum:
I also propose making testcase2 CO2 levels elevated to also test calculation of kpi above limit.

dhblum:
@JavierArroyoBastida I've addressed your comments, can you review again?

JavierArroyoBastida:
@dhblum this looks good to me. Please feel free to merge.

Issue #85, opened 2019-07-15
TestCase: BESTEST Air System


dhblum:
This issue is to develop a test case with a BESTEST building and air-based HVAC system model.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/161.

PR #161, opened 2020-03-19
Issue85 testcase bestest air

dhblum:
For #85.

Adds IBPSA Project 1 WP1.2 BESTEST air system test case model and associated data files.


dhblum:
@Mathadon @JavierArroyoBastida I think this test case is ready, or very close to ready, for merging to master.  Can you provide one more review before doing so?

JavierArroyoBastida:
@dhblum I've gone through the test case and I think this is ready to merge.

Mathadon:
go for it!

Issue #112, opened 2019-10-14
Add capability for browser access


dhblum:
This issue is to address a problem with accessing the API of the BOPTEST testcase from a browser (e.g. using javascript), which can be fixed by adding the Flask-Cors package (https://pypi.org/project/Flask-Cors/) to the REST API.  

@SenHuang19 is this correct?  Can you post the error you are getting?

SenHuang19:
Yes, the error that was showing in the console of a web browser is:  

> Access to XMLHttpRequest at 'http://127.0.0.1:5000/step' from origin 'null' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource

At the server side, the following log showed up as well:

> "OPTIONS /step HTTP/1.1" 404 - 

To mitigate the problem, I suggest to add the following code #114  to the restapi.py

> cors = CORS(app, resources={r"*": {"origins": "*"}})

This will allow the request from certain origins to access the server (by default it allows request from any domain but developers can modify it based on their needs)

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/167.

PR #167, opened 2020-03-30
Issue112 browser access

dhblum:
For #112.


dhblum:
@SenHuang19 the example for testcase2 is working ok, but for testcase 1 it seems to just be freezing when building the docker container with gecko.  Both locally and on travis ci.  At the lines:

```
Requirement already satisfied: six>=1.5 in /usr/lib/python2.7/dist-packages (from python-dateutil>=2.5.0->pandas)
Installing collected packages: numpy, pytz, python-dateutil, pandas
Successfully installed numpy-1.16.6 pandas-0.24.2 python-dateutil-2.8.1 pytz-2020.4
```

Anything that was done to testcase2 example that also needs to be done to testcase1?

dhblum:
I think the issue in the above comment (https://github.com/ibpsa/project1-boptest/pull/167#issuecomment-745625366) is just that testcase1 takes too long to run, because there are many control steps and at each step the ``/results`` and ``/kpi`` requests are called (in function ``send()`` in the html files).  It takes longer than the travis timeout time to complete the simulation.  I think only calling ``/results`` and ``/kpi`` at the end will solve the problem.

dhblum:
@SenHuang19 I proposed fixes in https://github.com/ibpsa/project1-boptest/pull/167/commits/05f71fd963fa7e13c039bf187a6ee69371db25e6 and https://github.com/ibpsa/project1-boptest/pull/167/commits/d381c56139401377d5753d5e2ac2dcd12efcb793 so that unit tests are passing and I think largely leave your code intact.  The main change is to only call ``/kpis`` and ``/results`` in the .html files only AFTER the simulation time has elapsed, and not at each step.  Please, can you review the changes and let me know if they are acceptable to you?  They achieve the ability to run the examples with the ``wrapper.py`` and report results in a timely fashion, however, I'm not sure if they affect other objectives you had with the code in terms of usage with a browser gui.

SenHuang19:
@dhblum Your modifications look good to me.  By the way, I created a new pull request which contains addtional modifications to js examples. Could you please review it?

dhblum:
This is ready to merge when tests pass.

Issue #157, opened 2020-03-16
Create testcase 3


JavierArroyoBastida:
This issue is to create a new testcase for unit-testing equivalent to testcases 1 and 2 but multi-zone. The purpose is to check multi-zone features in BOPTEST like the assignment of different names and set-points/thresholds to different zones. The creation of this new testcase has been triggered in https://github.com/ibpsa/project1-boptest/pull/133#issuecomment-598804397

JavierArroyoBastida:
@dhblum what do you think of using `Buildings.Examples.HydronicHeating.TwoRoomsWithStorage` for this test case?

dhblum:
Do you think we could accomplish the same with a multi-zone version of testcase1?  To keep it simpler for testing?  Or is it worth having a more complex case that you mention?

JavierArroyoBastida:
@dhblum maybe keeping it simple is the best for testing indeed. testcase2 already tests a certain level of complexity. I'll come with a proposal extrapolating testcase1 to multi-zone.

JavierArroyoBastida:
This is being addressed in https://github.com/JavierArroyoBastida/project1-boptest/tree/issue157_testcase3

JavierArroyoBastida:
@dhblum could you have a look at https://github.com/JavierArroyoBastida/project1-boptest/tree/issue157_testcase3? the model of the new testcase3 emulates two zones: north and south, each of them is exactly the same as the one of testcase1 and therefore have the same behavior. The idea is that each zone will differ only on their set points to stress that different KPI results will come from different set point readings. If you agree with this set up I'll work on the multi-zone specification unit tests.

dhblum:
@JavierArroyoBastida Yes I think this looks like a good initial test.  One quick change I would recommend is to change the ``zone`` parameters of the measurement blocks to be one word instead of two.  E.g. ``"North"`` and ``"South"``.  You may also make those names parameters of the model, and use those parameters in the ``zone`` parameter for each measurement tag, so the zone names are actually only defined in one place, instead of hard-coded separately in the ``TRooAir`` and ``CO2RooAir`` measurement blocks.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/169 via #121.

Issue #121, opened 2019-11-04
Error in specifying zone for data management


dhblum:
This issue is to address an error in the data manager when specifying specific zone for internal gains, occupancy, and setpoint schedules.  

For example, if the column name is ``LowerSetp[zon]``, then the following error is encountered when trying to compile the FMU with the data manager:

```
Traceback (most recent call last):
  File "compile_fmu.py", line 23, in <module>
    fmupath = compile_fmu()
  File "compile_fmu.py", line 18, in compile_fmu
    fmupath = parser.export_fmu(modelpath, [mopath])
  File "/home/dhbubu/git/ibpsa/project1-boptest/project1-boptest/parsing/parser.py", line 198, in export_fmu
    man.save_data_and_kpisjson(fmu_path=fmu_path)
  File "/home/dhbubu/git/ibpsa/project1-boptest/project1-boptest/data/data_manager.py", line 136, in save_data_and_kpisjson
    self._append_csv_data() 
  File "/home/dhbubu/git/ibpsa/project1-boptest/project1-boptest/data/data_manager.py", line 85, in _append_csv_data
    if appended[key] is not None:
KeyError: 'LowerSetp[zon]'
```

The fix for the error is to instead specify  the column name as ``LowerSetp``.  However, this does not allow for data specific to individual zones, which may be the case in multi-zone buildings.

The suggested fix is for the data manager to look for column names with ``LowerSetp`` in them, instead of explicitly with that name.  This may have consequences downstream with data retrieval that would have to be fixed as well.  @JavierArroyoBastida What do you think?

UPDATE [12/3/2019]:  Add to-do list based on @JavierArroyoBastida below comments and further discussion:
- [x] Add a parameter for specifying a zone designator to the read Modelica signal exchange block if KPI designators of ``AirZoneTemperature`` or ``OperativeZoneTemperature`` or ``RadiativeZoneTemperature`` are chosen.  UPDATE [2/23/2020]:  Being addressed in https://github.com/ibpsa/modelica-ibpsa/pull/1295.  UPDATE [3/4/2020]:  https://github.com/ibpsa/modelica-ibpsa/pull/1295 has been closed.
[...]

PR #169, opened 2020-04-05
Issue121 zone designation

dhblum:
This is for #121 and #157 and continues from https://github.com/ibpsa/project1-boptest/pull/133 for merge to master.


data/data_manager.py
```diff
@@ -67,29 +67,33 @@ def _append_csv_data(self):
...
-                        elif key not in all_keys:
-                            df.drop(key, inplace=True)
+                cols = df.keys()
+                if 'time' in cols:
+                    for col in cols.drop('time'):
+                        # Raise error if col already appended
+                        if col in appended.keys():
+                            raise ReferenceError('{0} in multiple files within the Resources folder. These are: {1}, and {2}'.format(col, appended[col], f))
+                        # Trim df from cols that are not in categories
+                        elif not( (col in all_keys) or any(col.startswith(key) for key in zon_keys) ):
```
dhblum, line 92: @JavierArroyoBastida Should there be a stronger check of the format ``'<category>[<zone>]'``?  In https://github.com/ibpsa/project1-boptest/pull/169/commits/26d89cf9d0d97985eaf034d9b7849e198e673d50 line 111-112 there is a check, but would be worth having when the test case is being compiled, instead of waiting until trying to calculate kpis.

kpis/kpi_calculator.py
```diff
@@ -100,27 +100,40 @@ def get_thermal_discomfort(self, plot=False):
...
-            tdis_dict[signal[:-1]+'dTupper_y'] = \
-                trapz(dT_upper,self.case.y_store['time'])/3600.
-            tdis_tot = tdis_tot + \
-                      tdis_dict[signal[:-1]+'dTlower_y'] + \
-                      tdis_dict[signal[:-1]+'dTupper_y']
+        
+        for source in self.case.kpi_json.keys():
+            if source.startswith(self.sources[0]):
+                # This is a potential source of thermal discomfort
+                zone_id = source.replace(self.sources[0]+'[','')[:-1]
```
dhblum, line 111: @JavierArroyoBastida Why change from referencing source by name to index of ``self.source``?  If by index, order of listing matters for future changes.

kpis/kpi_calculator.py
```diff
@@ -150,21 +163,32 @@ def get_iaq_discomfort(self, plot=False):
...
-            dI_upper[dI_upper<0]=0
-            idis_dict[signal[:-1]+'dIupper_y'] = \
-                trapz(dI_upper,self.case.y_store['time'])/3600.
-            idis_tot = idis_tot + \
-                      idis_dict[signal[:-1]+'dIupper_y']
+        
+        for source in self.case.kpi_json.keys():
+            if source.startswith(self.sources[4]):
+                # This is a potential source of iaq discomfort
+                zone_id = source.replace(self.sources[4]+'[','')[:-1]
```
dhblum, line 174: @JavierArroyoBastida Same as previous comment on index.

parsing/TwoZones.mo
```diff
@@ -0,0 +1,198 @@
...
+    annotation (Placement(transformation(extent={{30,60},{50,40}})));
+  Modelica.Thermal.HeatTransfer.Components.ThermalResistor resNor(R=0.01)
+    "Thermal resistance of north zone to outside"
+    annotation (Placement(transformation(extent={{0,50},{20,70}})));
+  Modelica.Thermal.HeatTransfer.Sensors.TemperatureSensor senTZonNor
+    "Room air temperature sensor of north zone"
+    annotation (Placement(transformation(extent={{60,50},{80,70}})));
+  Modelica.Blocks.Sources.Step setNor(
+    height=2,
+    offset=273.15 + 20,
```
dhblum, line 86: @JavierArroyoBastida Suggest change offset just to match setpoints.csv to 21 C instead of 20 C for north zone.

kpis/kpi_calculator.py
JavierArroyoBastida: @dhblum I myself was doubting on whether to do it with index or name. The benefit that I saw behind using the index was to have the source names in only one place so that changing the names wouldn't affect the rest of the code. But indeed, in this setting order of listing matters and may be less intuitive. I'll change it to refer source by name.

dhblum: I see what you're saying and agree there is a tradeoff.  Although, I think using the name makes the code more readable.  Also, we have the names separately listed in the Modelica IBPSA Library anyway.

data/data_manager.py
JavierArroyoBastida: @dhblum I agree it's a good idea to have the check when the test is being compiled. Shall I then remove the check in the KPI calculator? I cannot think of a case where the check in the KPI calculator could be useful if the check in the data manager has already filtered out the wrong keys.

dhblum: I would keep the check in the KPI calculator too, in the case the test case fmu is not compiled using the data manager.

data/data_generator.py
```diff
@@ -346,8 +346,8 @@ def generate_occupancy(self, occ_num,
         day_time_index = df.between_time(start_day_time, 
                                          end_day_time).index
 
-        df.loc[df.index.isin(day_time_index), 'Occupancy'] = occ_num
-        df.loc[~df.index.isin(day_time_index),'Occupancy'] = 0
+        df.loc[df.index.isin(day_time_index), 'Occupancy[1]'] = occ_num
```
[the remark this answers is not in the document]
JavierArroyoBastida: Addressed in https://github.com/ibpsa/project1-boptest/pull/175.

kpis/kpi_calculator.py
JavierArroyoBastida: Addressed in https://github.com/ibpsa/project1-boptest/pull/178.

data/data_manager.py
JavierArroyoBastida: What if we include the stronger check also when loading the test case with the data? so include it in:
https://github.com/ibpsa/project1-boptest/blob/f985196de852d8b833e2a56d12a8a11b5eae0650/data/data_manager.py#L247
This way the test case data will always undergo the test independently of how it was compiled.  In that case, wouldn't it be the checks at KPI Calculator redundant? We can always keep them but I think it's better if we can avoid them as KPI Calculator could be called iteratively if the user wants to use it on the fly.

dhblum: Yes, you're right, checking when loading the data would be best.  It should not be needed in the KPI Calc then.

dhblum:
@JavierArroyoBastida I've resolved conflicts and merged latest master.  I think this is ready to merge once checks pass.  Let me know if there if anything else.

dhblum:
I've also updated design requirements documentation in https://github.com/ibpsa/project1-boptest/pull/169/commits/43889e091a9668aacde252ed06a2226f30257dc3.

JavierArroyoBastida:
@dhblum that's great, thanks! I think this is ready to merge as well.

PR #173, opened 2020-04-07
modified the reset function 

SenHuang19:
This modification allows users to specify
1.  the start time 
2.  the warmup period length


dhblum:
This is the functionality we want, thanks!  Looking at it, I think we should abstract some of the common functionality between ``TestCase.__init__()``, your proposed ``TestCase.reset()``, and perhaps ``TestCase.advance()`` so as not to duplicate code.  Also need to add/correct unit tests as necessary.

Issue #179, opened 2020-04-10
Stronger check in data manager


JavierArroyoBastida:
This is to implement a stronger test in the data manager to check that zone keys from test case data are compliant with the zone identifier of the emulator model as suggested in https://github.com/ibpsa/project1-boptest/pull/169#discussion_r403736070.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/180.

PR #180, opened 2020-04-13
Issue179 data manager check

JavierArroyoBastida:
This is for https://github.com/ibpsa/project1-boptest/issues/179.

### DIFF
--- a/data/data_manager.py
+++ b/data/data_manager.py
@@ -56,7 +56,78 @@ class Data_Manager(object):
         # Load possible data keys
         with open(os.path.join(data_dir,'categories.json'),'r') as f:
             self.categories = json.loads(f.read()) 
+    
+    def _get_zone_identifiers(self):
+        '''Get all zone identifiers from the kpis.json file generated
+        by the parser. The zone identifiers can be used, for instance,
+        to check that the right data variables are saved in the 
+        compiled fmu or loaded from the fmu into the test case object. 
         
+        Returns
+        -------
+        zone_ids: list
+            List with all zone identifiers in the building model.
+        
+        
+        '''
+        
+        # Find the kpi.json
+        try:
+            # Case when we are loading data into a test case
+            kpi_json = self.case.kpi_json 
+        except:
+            # Case when we are saving data into a compiled fmu
+            with open(self.kpi_path,'r') as f:
+                kpi_json = json.loads(f.read()) 
+        
+        # Initialize list with zone identifiers
+        zone_ids = []
+        
+        # Go through kpi_json to find zone identifiers
+        for source in kpi_json.keys():
+            if '[' and ']' in source:
+                zone_id = source[source.find('[')+1:source.find(']')]
+                if zone_id not in zone_ids:
+                    zone_ids.append(zone_id)
+        
+        return zone_ids
+    
+    def _get_zone_and_boundary_keys(self):
+        '''This method returns all possible data keys differentiating
+        between zone-related keys and boundary data keys.  
+        The zone keys are keys containing a zone identifier such as
+        temperature set points, internal gains or occupancy. Boundary
+        keys are all the others, i.e. weather, emission factors, or 
+        prices:
+            all_keys = zon_keys + bou_keys
+            
+        Returns:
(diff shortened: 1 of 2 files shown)


data/data_manager.py
```diff
@@ -87,10 +152,19 @@ def _append_csv_data(self):
...
                         if col in appended.keys():
-                            raise ReferenceError('{0} in multiple files within the Resources folder. These are: {1}, and {2}'.format(col, appended[col], f))
-                        # Trim df from cols that are not in categories
-                        elif not( (col in all_keys) or any(col.startswith(key) for key in zon_keys) ):
+                            raise ReferenceError('{0} in multiple files within the Resources folder.'\
+                                                 'These are: {1}, and {2}'.format(col, appended[col], f))
+                        # Trim df from cols that are not allowed
+                        elif not( (col in zon_keys) or (col in bou_keys) ):
                             df.drop(col, inplace=True)
+                            raise KeyError('The data column {0} from file {1} '\
```
dhblum, line 160: @JavierArroyoBastida Raising an error in ``load_data_and_kpisjson()`` is good since that is the data is being used in the deployed test case.  For ``_append_csv_data()``, in the past a column that did not conform to the specification was just dropped and not included in the final test case fmu.  With this implementation, an error is being raised and the process is being stopped if this is the case.  Just want to double check that this functionality is what you intend.  It is a stricter test, though somewhat less convenient if a column in a csv used for test data is derived from other columns in a csv, which may not be used for test data.  I am tending toward the stricter test, but wanted to double check with you.

JavierArroyoBastida: @dhblum I was indeed doubting whether a warning or an error was most convenient. In the end I decided to use the error because it avoids saving (and loading) the test case with unnecessary data. If these data needs to be saved for any reason it can still be stored in any other folder. Also if the test case developer is not using the `data_generator` but excel sheets with dependent columns, it should not be an issue to filter out the dispensable columns, for instance when changing the format from xlsx to csv.

dhblum: Just to be clear 

> In the end I decided to use the error because it avoids saving (and loading) the test case with unnecessary data

refers to saving and loading data from the test case DIRECTORY, not the test case FMU?  Because the unused data is dropped before saving the test case FMU.

JavierArroyoBastida: By "saving" I refer to the process of storing the data from the test case directory into the fmu like in:
https://github.com/JavierArroyoBastida/project1-boptest/blob/4b37fbcdeff844878e10310722a0c200f0733fc9/data/data_manager.py#L183
By "loading" I refer to the process of retrieving the data from the fmu into the test case object like in:
https://github.com/JavierArroyoBastida/project1-boptest/blob/4b37fbcdeff844878e10310722a0c200f0733fc9/data/data_manager.py#L327
Both methods are implemented with errors instead of warnings to prevent both: saving unnecessary data into the test case fmu and loading unnecessary data into the test case object. The latter would be used only for compiled tests case fmus that save their data by other means than using the `data_manager.save_data_and_kpisjson` method, would there be any case like that one? I'm wondering now if it's really needed...

dhblum: Yes, a test case developer could compile their own FMU with data.

And ok about the "saving" and "loading".  My point is that ``save_data_and_kpisjson `` uses ``_append_csv_data()``.  If the "saving" method uses a warning rather than an error, than the unnecessary data would still NOT be saved in the test case FMU in the end.  It is even described by https://github.com/JavierArroyoBastida/project1-boptest/blob/4b37fbcdeff844878e10310722a0c200f0733fc9/data/data_manager.py#L170.

JavierArroyoBastida: You're right, I didn't notice but see your point now. Still I think it's better practice to raise an error to ensure that emulator developers know what is being saved in the test case fmu, warnings can be easily missed.

dhblum: That sounds good.  So should Line 170 - 176 and 159 be edited?

JavierArroyoBastida: Indeed, I'm going to push changes for that.

Issue #181, opened 2020-04-14
Change testcase3 offset


JavierArroyoBastida:
This is to change the testcase3 offset in the test case modelica model to coincide with the set-point used in the testcase setpoints.csv file as suggested by https://github.com/ibpsa/project1-boptest/pull/169#discussion_r403739171.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/182.

PR #182, opened 2020-04-14
Issue181 change testcase3 offset 

JavierArroyoBastida:
To address https://github.com/ibpsa/project1-boptest/issues/181.


dhblum:
I will push changes to correct unittest.

PR #187, opened 2020-04-18
Issue117 reset time

dhblum:
For #117 

- ``/reset`` API now takes two arguments: "start_time" and "warmup_time"
- abstracted common code
- code formatting
- update unittests and examples


dhblum:
@SenHuang19 I made a few changes from your original PR:

- added documentation to new code
- re-formatted some code
- changed the examples to start at 0 (as they were originally) and instead improved upon the unit test for ``/reset`` to test that the functionality works ok

Please review and let me know if you have additional comments or feedback.  It is close to merging I think after checks pass.

SenHuang19:
@dhblum Your changes look good to me. Just one question: do we need to allow users to specify the input_object during the warm-up period?

dhblum:
@SenHuang19 My first thought there is then the warmup period becomes very similar to using the ``/advance`` request, except that at the end the data storage is reset.  Perhaps it is easier to just add an API call to reset the data storage?  Then, a user can combine the API requests into the functionality as they wish?

I've also made some fixes and waiting for checks to pass.

Also, FYI @JavierArroyoBastida as this PR changes the API of the ``/reset`` request.

dhblum:
@SenHuang19 Based on our last discussion, how about if we change the api name from "reset" to "initialize", which is more clear that it is initializing the test case to a specified starting time with a warm-up period?

SenHuang19:
Sounds good to me.

dhblum:
@SenHuang19 I've updated the API.  Once checks pass, I think this is ready to merge.  Do you have any more comments on it?

dhblum:
I suggest merging https://github.com/ibpsa/project1-boptest/pull/189 first.

SenHuang19:
It looks good to me and I have no further comments. Good job!

Issue #188, opened 2020-04-21
Clean up unittests for test cases


dhblum:
This issue is to clean up the process of unit testing for test cases.  Currently, there is quite a bit of manual specification of reference results directly in Python.  This issue will make greater use of automatic generation and subsequent comparison of reference files.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/189.

PR #189, opened 2020-04-21
Issue188 clean test case unittests

dhblum:
For #188 
- Standard API tests in ``partialTestAPI`` auto-generate and compare reference results as files in dedicated directories for given test case.
- Added methods to ``testing/utilities.py`` for checking json files and key-value pairs
- Updated ``testing/test_testcase1.py``, ``testing/test_testcase2.py``, and ``testing/test_testcase3.py``as necessary, removing manually-defined reference results.


dhblum:
@JavierArroyoBastida Can you please review when you get a chance?  This will make setting up unit tests for new test cases much easier in my opinion.

testing/utilities.py
```diff
@@ -105,6 +105,69 @@ def compare_ref_timeseries_df(self, df, ref_filepath):
             
         return None
     
+    def compare_ref_json(self, json_test, ref_filepath):
+            '''Compare a json to a reference json saved as txt.
```
JavierArroyoBastida, line 109: Is the ".txt" extension used for any reason? So far we've used ".json" which I think was more clear.

testing/utilities.py
```diff
@@ -235,46 +322,40 @@ def test_get_inputs(self):
...
         
         inputs = requests.get('{0}/inputs'.format(self.url)).json()
-        self.assertEqual(len(inputs), len(self.inputs_ref))
-        for inp in inputs:
-            self.assertTrue(inp in self.inputs_ref)
-            self.assertTrue(inputs[inp]['Unit'] == self.inputs_ref[inp]['Unit'])
-            self.assertTrue(inputs[inp]['Description'] == self.inputs_ref[inp]['Description'])
-            self.assertTrue(inputs[inp]['Minimum'] == self.inputs_ref[inp]['Minimum'])
-            self.assertTrue(inputs[inp]['Maximum'] == self.inputs_ref[inp]['Maximum'])
+        ref_filepath = os.path.join(get_root_path(), 'testing', 'references', self.name, 'get_inputs.txt')
```
JavierArroyoBastida, line 325: This reference file is missing for each test case.

testing/utilities.py
```diff
@@ -235,46 +322,40 @@ def test_get_inputs(self):
...
 
         measurements = requests.get('{0}/measurements'.format(self.url)).json()
-        self.assertEqual(len(measurements), len(self.measurements_ref))
-        for measurement in measurements:
-            self.assertTrue(measurement in self.measurements_ref)
-            self.assertTrue(measurements[measurement]['Unit'] == self.measurements_ref[measurement]['Unit'])
-            self.assertTrue(measurements[measurement]['Description'] == self.measurements_ref[measurement]['Description'])
-            self.assertTrue(measurements[measurement]['Minimum'] == self.measurements_ref[measurement]['Minimum'])
-            self.assertTrue(measurements[measurement]['Maximum'] == self.measurements_ref[measurement]['Maximum'])
+        ref_filepath = os.path.join(get_root_path(), 'testing', 'references', self.name, 'get_measurements.txt')
```
JavierArroyoBastida, line 334: Same here, I think these reference files are missing.

testing/utilities.py
dhblum: Sounds good, changed in https://github.com/ibpsa/project1-boptest/pull/189/commits/b76cfb448de10d6b64a91e7f497c6de96709156c.

testing/utilities.py
dhblum: Good catch.  This is because the ``.gitignore`` has a line *.txt, and I did not notice that those files did not get staged.  Changing to .json as suggested above fixed this problem and these files have been committed in https://github.com/ibpsa/project1-boptest/pull/189/commits/b76cfb448de10d6b64a91e7f497c6de96709156c.

testing/utilities.py
dhblum: Same issue as previous comment.  Fixed in https://github.com/ibpsa/project1-boptest/pull/189/commits/b76cfb448de10d6b64a91e7f497c6de96709156c.

dhblum:
@JavierArroyoBastida Thanks for the comments, I addressed them.  Regarding the KPI calculator testing, it would be good to utilize the new methods for comparison to be consistent.

dhblum:
@JavierArroyoBastida I've edited ``test_kpis.py`` and updated reference results.  Checks are still running, but can you take a second look before merging, assuming checks pass?

JavierArroyoBastida:
@dhblum thanks for the edit in `test_kpis.py`. I started up a parallel implementation but this one looks more convenient. I believe this pull request is ready after https://github.com/ibpsa/project1-boptest/pull/197 is merged.

Issue #163, opened 2020-03-20
TestCase: Multizone Residential Hydronic


dhblum:
This issue is to add the Multizone Residential Hydronic test case from IBPSA Project 1 WP1.2.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/192.

PR #192, opened 2020-04-21
Issue163 testcase multi zone residential hydronic

JavierArroyoBastida:
This is for https://github.com/ibpsa/project1-boptest/issues/163. Changes are still required to update to latest BOPTEST features like the specification of zone identifyiers for multi-zone buildings.


JavierArroyoBastida:
@dhblum @valgavan this emulator is about to be ready, but there are still missing tasks like:

- [ ] Add CO2 taggs.
- [ ] Clean up model from unnecessary files, particularly of bak.mo files that came along with the model. 
- [ ] Organize documentation. Right now the documentation is a bunch of files. Would be nice to make a selection and write it down in a unique location gathering the most relevant info. 
- [ ] Improve unit tests adding, for instance, references to KPI calculation. May be useful to merge [issue188](https://github.com/ibpsa/project1-boptest/pull/189) for this. 
- [ ] Go through the checklist once again when all tasks are accomplished. The model has experienced little changes in the BOPTEST integration process, however compilation does not work the same in JModelica than Dymola, so particularly interesting would be to make sure that the emulator runs for a year in BOPTEST. Also would be nice to have the peer review checklist stored with the model to keep track of the comments of the reviewer.  

I'm particularly concerned about commit https://github.com/ibpsa/project1-boptest/pull/192/commits/8c131bf652caed0da1a0ea49a2a29543ae3a7c4d. In that commit I had to remove the definition of the window overhang geometry because that was affecting somehow the loading of the weather data into the FMU. @dhblum do you have any idea why that could happen? I don't see the relation at all, but if you checkout the previous commit (https://github.com/ibpsa/project1-boptest/pull/192/commits/d6901d230afcca2558b81a1866edf82ac6ccf50e) and you do `make compile_testcase_model TESTCASE=multizone_residential_hydronic` you'll realize that the model is compiled but that the weather file is not loaded... Maybe something to be reported in the Buildings library? Also, @valgavan, do you think the window overhang could have a significant influence? If yes we'll have to think how to include it while keeping the weather file within the FMU.

dhblum:
The issue with the overhang and weatherfile you mention is a bit strange.  I compiled ``Buildings.ThermalZones.Detailed.Examples.MixedAirFreeResponse`` which has the same overhang definition and the weather file was in the FMU.  Will need to look into this more.

JavierArroyoBastida:
Indeed, very strange. Would be nice if you could try to reproduce by checking out commit (https://github.com/ibpsa/project1-boptest/pull/192/commits/d6901d230afcca2558b81a1866edf82ac6ccf50e) and doing `make compile_testcase_model TESTCASE=multizone_residential_hydronic`. Maybe it is related to my setup but I wouldn't expect it because most of the process runs in the container. I've tried with both, the master and the v6.0.0 versions of Buildings and obtained same results.

JavierArroyoBastida:
@dhblum, @valgavan, @TaoYang-CFEI, I've gone through all points in https://github.com/ibpsa/project1-boptest/pull/192/#issuecomment-620532780 but the last one. Last missing point is basically a peer review using latest version of the peer review checklist.

dhblum:
To confirm from the conversation in https://github.com/ibpsa/project1-boptest/pull/192,  @TaoYang-CFEI will review this model with the latest checklist, v1.3.

TaoYang-CFEI:
@JavierArroyoBastida @dhblum Please find attached the updated review checklist, please let me know if there is anything else I am missing.
[BOPTEST Test Case Peer Review Document_draftv1_3_Engie_emulator.docx](https://github.com/ibpsa/project1-boptest/files/5168148/BOPTEST.Test.Case.Peer.Review.Document_draftv1_3_Engie_emulator.docx)

JavierArroyoBastida:
@TaoYang-CFEI thank you very much for the review! Could you please be more specific? which model assumptions are not documented? Which inputs/outputs need further description? that would help a lot!

TaoYang-CFEI:
@JavierArroyoBastida For the model inputs/outputs, I think variable names, unit, max, min, and description should be added for each I/O, which is blank in the current documentation. Also, it would be nice if there is a subsection "other assumptions" collecting some other unmentioned assumptions, such as ground temperature. David sent me an example of documentation, which may be helpful to take a look. [https://github.com/ibpsa/project1-boptest/blob/master/testcases/bestest_air/doc/index.html#L226](url)

JavierArroyoBastida:
@TaoYang-CFEI thank you very much for your comments. I've addressed your requests within this PR. Please let me know if something is missing.

dhblum:
There are conflicts that need to be fixed.  Other than that, anything else pending before final review for merging?

JavierArroyoBastida:
Merge conflicts are now resolved and the model has been updated with the weather station. All unit tests are passing locally. This should be ready to merge after one last review.

dhblum:
Wow thanks a lot @JavierArroyoBastida.  I will try to review this by the end of the week.

dhblum:
I sent a BOPTEST peer review checklist v1.3 to @JavierArroyoBastida.

JavierArroyoBastida:
@dhblum @valgavan @TaoYang-CFEI, I've been working on improvements to address the issues raised by @dhblum in his last review. Particularly, these improvements are:
1. Addition of zone names to input signal descriptions. 
2. Addition of CO2 generation and read blocks for CO2 concentration. Still no controllability of any kind is added for ventilation. 
3. Addition of zone names to output signal descriptions for zone air temperature. 
4. Update to latest version Buildings and IBPSA masters. 
5. Fix image that was not being displayed in documentation. 
6. Addition of description of cooling control.
7. Fix cooling efficiency. 
8. Better description of ventilation in documentation.
9. Complete unfinished sentence in documentation.
10. Do not measure convective gains for consistency. 
11. Check that model runs for full year simulation in BOPTEST Docker and in Dymola with cvode and tol=1e-6. Notice that the control parameters (i.e. all proportional and integral constants from the numerous PI blocks) have been tuned to prevent the integration to stall. The control logic has been proven to be quite sensitive to these parameters which are very difficult to tune because of the strongly connected algebraic loops that lead the integration to stall quite randomly for different parameter choices. The most prominent algebraic loop happens for the heating system and involves the air indoor temperature, emission system supply water temperature, and emission system pump mass flow rate. The newly set of control parameters leads to a more robust behavior allowing to run the yearly simulation without stalling, but their change has affected the baseline performance as can be appreciated in https://github.com/ibpsa/project1-boptest/pull/192/commits/fdf73541a1ee37f01a4c04f7b323fc79e81958fa. Up to this commit any update of references was either for inclusion of CO2 read blocks or for small numerical differences. Still the baseline performance seems realistic to me, also with the new set of control parameters. 
12. Explain seasonal cooling mode in documentation. 

Additionally:
- Tests have been extended with dynamic and highly dynamic pricing.
- I/O descriptions have been updated in documentation.
- A general clean-up has been performed attemping to improve model readability.
- API has been updated according to latest changes in BOPTEST master. 
- Pre-commit checks have been passed. 

@dhblum could you review for merging once again?

dhblum:
Thanks a lot @JavierArroyoBastida for all of these updates and responding to my comments.  I will have a review, but probably won't complete until next week.

JavierArroyoBastida:
@dhblum, @valgavan, I've updated the hydronic parameters in https://github.com/ibpsa/project1-boptest/pull/192/commits/80e586801a260163d680d1fd75d7708a0b583b3f following our discussion by email. Unfortunately, the simulation keeps stalling with the new set of control parameters, so I had to change back to the previous values here: https://github.com/ibpsa/project1-boptest/pull/192/commits/5d87dc4eb016b95ace2a2c213b6aa5b15dd930a0. That makes me think that the problem with the siulation robustness is not related to physics but to numerics related to the control. @dhblum  could you simulate for a full year in Dymola  with the edits that you proposed by email? Maybe I'm just missing something. 

NOTE: References still need to be updated for whatever set of parameters that we end up using.

dhblum:
@JavierArroyoBastida Ok thanks, I will have a closer look.  I did not try for the full year.

dhblum:
I made updates in https://github.com/ibpsa/project1-boptest/pull/192/commits/19450f42870577016ddf3458ecbb56c937bb7761 and simulated with the following:

- Buildings Library commit: e48dd1bd7ba38aa40f9b4d886da3e616a86a0deb
- IBPSA Library commit: 8c3355524f0847a91021df2164076f23796441f0
- Dymola 2021 with cvode tol 1e-6 for 1 yr: OK
- JModelica Trunk from June 2019 with cvode tol 1e-6 for 1 yr: OK

JavierArroyoBastida:
I still don't get successful full year simulation when using Dymola **2020** with the new set of dp and control parameters, even when using the same solver and library versions. The only difference is the Dymola version, so it's probably worth the update to 2021... In JModelica it worked fine to me too, so I suggest we go with this new set of parameters. I've updated library versions https://github.com/ibpsa/project1-boptest/pull/192/commits/51ed6122082b6c43e26d44cbec277a43ec331de7, the compiled fmu https://github.com/ibpsa/project1-boptest/pull/192/commits/d01c0d0beb7b0a09a45d55f369f60752d505b0a9, and references https://github.com/ibpsa/project1-boptest/pull/192/commits/25afa6ee321a2558b8fa0cd9d1e142274016f288, so should be ready to merge once tests pass if I'm not missing something.

dhblum:
Great, thanks so much.  The one thing is Michael suggested increasing the two-way valve dp to 6 kPa from 1 kPa to give it better authority against the 6 kPa fixed dp.  I will test today and if all looks good I'll push the changes.

JavierArroyoBastida:
Ok great, thanks!

dhblum:
The change to 6 kPa looks good.  Just waiting for 1 yr simulation to finish with JModelica.

I hate to change things more, but wanted to ask anyway.  Does it makes sense that there are output variables that measure heat output for each radiator?  This would assume there is a heat meter installed across every radiator, which may not be realistic in a house.  A more realistic measurement variable may be the radiator valve position, especially since it is assumed that it can be overwritten.  For instance, there are smart trvs by [honeywell](https://heatingcontrols.honeywellhome.com/products/thermostats/thermostatic-radiator-valves/Wirelessly-Controlled-TRVs/HR92-Radiator-Controller/) and [danfoss](https://www.danfoss.com/en/products/smart-heating/dhs/smart-heating/danfoss-ally/#tab-overview). Infact, a [danfoss case study](https://www.danfoss.com/en/service-and-support/case-studies/dhs/ally-ballan-mir%C3%A9-case-story/) I found is for a house in France.

mwetter:
@dhblum : Can you please also make a PR in the Buildings library to change this on `Buildings.Examples.HydronicHeating.TwoRoomsWithStorage`.

dhblum:
@mwetter Yes, will do.

JavierArroyoBastida:
Better to change things earlier rather than later :)
In fact, heat meters across every radiator is not a common practice, although heat allocators are. See [here](http://www.ecothermo.it/heat-cost-allocator-drawbacks/) for a quite informative post on that topic. Heat allocators are not intrussive, hence you can find them in every radiation of a residential building. The drawback is that they are not as accurate. Still I agree that not reading any heating at all is more representative of current practice. I guess this relates to our discussion yesterday (and to our everlasting dilemma) on whether we want the user to only access measurements that are taken in current practice or to let the user access to a more "ambitious" set of measurements that could lead and justify further best practices. I'm more in favor of the latter, although also understand the point of the former. 
In any case, I think it's a good idea to read valve position in this particular case.

dhblum:
I don't know why the latest commit is failing.  Seems to be something with weather data (NTot).  But, the results were newly generated and the tests still pass on my machine...

dhblum:
Pushed a few more changes:
- More corrections for using Buildings Library only
- Added heating valve control signals as outputs
- Changed cooling overwrite to control signal between 0-1 instead of absolute cooling thermal power
- Edited names/descriptions of some points for clarity

I think it's really close.  I think the list is down to:

- [ ] Double check boundary condition data at event times as in https://github.com/ibpsa/project1-boptest/issues/309
- [ ] Double check documentation
- [ ] Updated wrapped.fmu
- [ ] Merge latest master to double check

Anything else?

JavierArroyoBastida:
@dhblum thanks a lot for your latest changes. I've already checked  boundary condition data at event times as in https://github.com/ibpsa/project1-boptest/issues/309 and updated references for that. I think not much is missing apart from merging master and double checking.

dhblum:
@JavierArroyoBastida Can you please merge the latest ibpsa master to the development branch for this PR?  Since it is on your fork, not sure if there is a specific way you handle it.  

Added to to-do that we need to add scenarios.  I can do that after you merge the latest master as simple commits.  Also have to figure out why the ci is failing, with an inability to connect to the test case for the tests.  Only this one, not the others though.

JavierArroyoBastida:
Done! I'm not sure why would that happen either... Will let you know if I figure out something.

dhblum:
The test wouldn't pass because the test case wouldn't deploy because the config.py wasn't configured correctly with the name and scenario.  Fixed and adding scenarios, etc.

testcases/multizone_residential_hydronic/doc/MultiZoneResidentialHydronic.html
```diff
@@ -0,0 +1,658 @@
...
+dwellings. Its area is approximately 120 m&sup2; (<i>S_ref</i>).
+The following figure shows the building layout and
+a sketch of the hydraulic system. The coloured elements in the scheme
+represent the controllable components
+through the BOPTEST interface. The dimensions are provided in metres.
+
+</p>
+<p align=\"center\">
+<img alt=\"Simulated residential dwelling\"
+src="images/layout.png\">
```
JavierArroyoBastida, line 22: @dhblum I can't see the image unless:
```suggestion
src="images/layout.png">
```
Is it the same for you?

testcases/multizone_residential_hydronic/doc/MultiZoneResidentialHydronic.html
```diff
@@ -0,0 +1,658 @@
...
+</p>
+
+<p>
+The orientation of each building was chosen to maximize the natural light during winter.
```
JavierArroyoBastida, line 26:
```suggestion
The orientation of the building was chosen to maximize the natural light during winter.
```

testcases/multizone_residential_hydronic/models/find_days.py
```diff
@@ -0,0 +1,14 @@
+'''
+Finds the peak and typical heat days for the multizone_residential_hydronic.
+This case needs to be deployed if using BOPTEST image.
+
+'''
+
+from data.find_days import find_days
+import json
+
+days = find_days(heat='sum_Q_flow', cool=None,
```
JavierArroyoBastida, line 10: @dhblum we leave out cooling because it is idealized right?

JavierArroyoBastida (commented):
@dhblum I've been reviewing and I think it's looking good in general. One question I have is: do we want to add the .docx files from the doc folder? Maybe we should remove them for consistency since other test cases do not include peer-review documentation.

testcases/multizone_residential_hydronic/doc/MultiZoneResidentialHydronic.html
dhblum: It works either way for me in Firefox on Ubuntu 18.04.  Let's take your suggestion since it seems to work better on your setup.

testcases/multizone_residential_hydronic/models/find_days.py
dhblum: Yes that was my thinking.  Does that make sense?  Maybe we discuss what to do with cooling in a future WP1.2 meeting?

dhblum:
Thanks for reviewing.  About the review docs, I agree with your suggestion to remove to be consistent with other test cases, but then let's discuss as a WP1.2 specifically what to do with them for all test cases.

testcases/multizone_residential_hydronic/models/find_days.py
JavierArroyoBastida: I think that makes sense. Still may be worth discussing this in a WP1.2 meeting.

JavierArroyoBastida:
Ok!

testcases/multizone_residential_hydronic/models/find_days.py
dhblum: Ok yes.

dhblum (approved):
I think this is ready to merge once tests pass one last time.

Issue #113, opened 2019-10-16
Add OperativeZoneTemperature for thermal discomfort KPI calculation


dhblum:
The KPI calculator uses measurement outputs with KPI ID "AirZoneTemperature" to calculate the thermal discomfort core KPI.  However, for zones served by radiant systems, it may be more appropriate to calculate thermal discomfort with the KPI ID "OperativeZoneTemperature."  A proposed solution is to edit line 110 of ``kpis/kpi_calculator.py`` (``for signal in self.case.kpi_json['AirZoneTemperature']:``) to include signals from ``self.case.kpi_json['OperativeZoneTemperature']`` as well.

JavierArroyoBastida:
@dhblum @icupeiro my suggestion is to include both, "AirZoneTemperature" and "OperativeZoneTemperature", and give the user the possibility to choose which one to use for comfort KPI calculation.

icupeiro:
I agree with @JavierArroyoBastida suggestion

In addition to @dhblum comment in #186 and thinking about the multizone-commercial-hydronic case, indeed the air system controls the supply air temperature, however being an hybrid GEOTABS system I find more appropiate to use the operative temperature metric (since the whole structure is activated and roof/walls effects is significant). Moreover, thermostats are usually placed against walls, so their measurement is affected to some extent by the radiant wall temperature.

dhblum:
As described in my initial description of this issue, I believe the only change necessary would be in ``kpis/kpi_calculator.py`` where instead of just searching for ``"AirZoneTemperature"``, ``"OperativeZoneTemperature"`` is also included.  The signal exchange read block already has an option for Operative Temperature and the parser will include it in the ``kpis.json`` just like Air Temperature.

PR #198, opened 2020-04-30
Issue113 Compute discomfort also for operative zone temperature readings

JavierArroyoBastida:
This is for https://github.com/ibpsa/project1-boptest/issues/113.


dhblum:
We should create a test for this and also make sure any documentation is updated.

JavierArroyoBastida:
@dhblum do you think we should implement a new testcase for unit-testing of this or use one of the existing?

dhblum:
To limit maintenance of an additional test case, would it make sense to change one of the zones in testcase3 to measure operative temperature?

JavierArroyoBastida:
That does make sense. It's actually what I had in mind. I'll take care of the implementation.

JavierArroyoBastida:
@dhblum I think this is ready to merge once tests pass. Could you please check?

Issue #204, opened 2020-05-18
Allow test case data to be "NA"


dhblum:
Currently, all columns of test case data need to be given a numeric.  However, an emulator may not utilize district heating energy or biomass, for instance.  Therefore, it may be more clear to indicate their prices or emission factors as NA, rather than 0.

Using "NA" in the test case data column for constant electricity price in test case 2:

- Using the ``/forecast`` API request returns ``"PriceElectricPowerConstant": [NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN]``.  This is ok in my opinion, as it indicates that the particular forecast value is not available in the given test case data.  During development and unit testing, this can be identified and checked for.

- Using the ``/kpi`` API request returns ``{"tdis_tot": 0.2873674410761013, "idis_tot": 0.0, "ener_tot": 5.1671772633773205, "cost_tot": NaN, "emis_tot": 2.5835886316886603, "time_rat": 0.07108319024244944}``.  This is ok in my opinion, as it indicates that the particular KPI could not be calculated with the given test case data.  During development and unit testing, this can be identified and fixed.

dhblum:
Correction to the description of this issue.  If a specific column of test case data is not defined, it is then not used within BOPTEST without much issue.  The variable is not delivered to the user as part of a forecast and as long as the data is not needed for KPI calculation, the KPI calculator does not look for it.  If it is needed for KPI calculation, then a KeyError will be thrown in the KPI calculator, which would be caught with unit testing and corrected.  

Therefore, if a test case emulator does not, say use Gas Power, then the developer can simply omit the specification of EmissionsGasPower and PriceGasPower, rather than specify "NA" in those columns and require handling of what "NA" means within the BOPTEST framework.

PR #205, opened 2020-05-18
Issue204 na

dhblum:
For #204


dhblum:
Closing as may not be necessary.

PR #208, opened 2020-05-19
add logging and messages to users

SenHuang19:


restapi.py
```diff
@@ -55,100 +73,302 @@ class Advance(Resource):
...
+        
+        try:
+        
+            step = case.get_step()
+            
+        except Exception as e:
+        
+            app.logger.error("Fail to return the simulation step:{}".format(e))
+        
+            return {'message':'failure','error':e,'result':None}
```
dhblum, line 165: Should 'message', 'error', 'result', be standard and ``None`` given as value where applicable, as is done in this line?  Other returns in proposed code have only a subset of these keys.

testcase.py
```diff
@@ -361,11 +372,22 @@ def set_forecast_parameters(self,horizon,interval):
...
+        
+             self.horizon = float(horizon)
         
-        self.horizon = float(horizon)
-        self.interval = float(interval)
+             self.interval = float(interval)
+            
+        except TypeError:
+        
+             return {'message':'failure','error':'insufficient input','result':None}
```
dhblum, line 384: It is not clear what the problem is with the input.  A better description is needed.  I don't think ``'insufficient'`` is the correct word here.

testcase.py
```diff
@@ -361,11 +372,22 @@ def set_forecast_parameters(self,horizon,interval):
...
+             
+        except ValueError:
         
-        return None
+             return {'message':'failure','error':'invalid input','result':None}
```
dhblum, line 388: It is not clear what the problem is with the input. A better description is needed. I don't think ``'invalid'`` is the correct word here.  It aligns more with ``TypeError``.

dhblum:
Closed in favor of https://github.com/ibpsa/project1-boptest/pull/272.

Issue #222, opened 2020-07-21
Set library versions json


JavierArroyoBastida:
This issue is to avoid the use of `set_library_versions.sh` as permission errors arise when these files are not originally generated in a Linux machine (independently of their content). See for instance:

https://travis-ci.com/github/ibpsa/project1-boptest/builds/176472241 
or
https://travis-ci.com/github/ibpsa/project1-boptest/builds/176475619 

An alternative may be to store the versions in a json file and checkout the libraries from a python script that reads such file.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/223.

PR #223, opened 2020-07-21
Issue222 Use json file and python script to set library versions.

JavierArroyoBastida:
This is for https://github.com/ibpsa/project1-boptest/issues/222.


testing/set_library_versions.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+"""
+This file is used by unit testing to specify library versions.
+Commands here should maintain alignment with the environment generated in 
+testing/Dockerfile.
```
dhblum, line 5: Add documentation for argument.

testing/set_library_versions.py
```diff
@@ -0,0 +1,24 @@
...
+with open(os.path.join(sys.argv[1],'models','library_versions.json'),'r') as f:
+    library_versions = json.loads(f.read()) 
+
+# Checkout specified commit for each library
+for key in library_versions.keys():
+    if key == "IBPSA_COMMIT":
+        os.system('cd $MODELICAPATH/IBPSA && git checkout {}'.format(library_versions[key]))
+    elif key == "BUILDINGS_COMMIT":
+        os.system('cd $MODELICAPATH/Buildings && git checkout {}'.format(library_versions[key]))
+    elif key == "IDEAS_COMMIT":
```
dhblum, line 23: Suggest raise error if key not found with message saying so.

dhblum:
@Mathadon This replaces the set_library_versions.sh script that we had discussed before.  Just wanted to check if you had any comments on this implementation instead.  It allows for the spec of a plain json in the test case model directory instead of actually adding a shell script and then running it during unittesting.

testing/set_library_versions.py
```diff
@@ -0,0 +1,24 @@
...
+import os
+
+# Read library versions from testcase library_versions.json file
+with open(os.path.join(sys.argv[1],'models','library_versions.json'),'r') as f:
+    library_versions = json.loads(f.read()) 
+
+# Checkout specified commit for each library
+for key in library_versions.keys():
+    if key == "IBPSA_COMMIT":
+        os.system('cd $MODELICAPATH/IBPSA && git checkout {}'.format(library_versions[key]))
```
Mathadon, line 20: To avoid code injection you can add double quotes around the curly braces, see https://stackoverflow.com/questions/15783701/which-characters-need-to-be-escaped-when-using-bash

testing/set_library_versions.py
JavierArroyoBastida: Addressed in https://github.com/ibpsa/project1-boptest/pull/223/commits/34af1684c40bb7e1eb982103ef235afa5ce636d0 .

testing/set_library_versions.py
JavierArroyoBastida: Addressed in https://github.com/ibpsa/project1-boptest/pull/223/commits/237ad498ac225b24530be09eabeba42b62a7cdc2.

testing/set_library_versions.py
JavierArroyoBastida: Sorry I'm not sure what you mean from the link. Can you quckly type what you have in mind? no need to be perfect, just to see the idea.

dhblum: I think use:
```
os.system('cd $MODELICAPATH/IBPSA && git checkout "{}"'.format(library_versions[key]))
```
so that the contents of the bracket is always a string argument of ``git checkout`` rather than potentially unwanted additional code and commands.

Mathadon: yup

JavierArroyoBastida: aha, I see now. Implemented here https://github.com/ibpsa/project1-boptest/pull/223/commits/c50e790076479a8ac0dddc6ddaabb1af5434b741.

dhblum:
@Mathadon Just checking if you approve the change and have no other comments, I can merge.

Mathadon:
no further comments!

Issue #224, opened 2020-07-22
Initialize return


JavierArroyoBastida:
@dhblum, @SenHuang19, is there a reason why `testcase.initialize()` is returning a boolean? see: https://github.com/ibpsa/project1-boptest/blob/042ef3965f6c9ea18e2ce7bd69901cabdc69d572/testcase.py#L239
From personal use I'd find much more useful that it returns the same format as a regular `testcase.advance()` call to use it as an initial measurement for starting up the control algorithms. What do you think? if you agree I can work on that.

dhblum:
I think that's a nice catch and sounds good.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/225.

PR #225, opened 2020-07-23
Issue224 Initialize return

JavierArroyoBastida:
For https://github.com/ibpsa/project1-boptest/issues/224.


dhblum:
This looks good to me.  @SenHuang19 Any objections to merging?

SenHuang19:
> This looks good to me. @SenHuang19 Any objections to merging?

I agree that this looks very good. Just one comment: do we need to add a class variable to the testcase which indicates the initialization is done successfully or not. In this case, we check if this variable in return from the "initialize_result" is true or false rather than if the return contains "time"

dhblum:
@SenHuang19 I'm not sure I understand your comment.  ``initialize`` returns the measurement dictionary if successful, otherwise it returns None.  A controller can then check if None is returned to determine if successful.

JavierArroyoBastida:
@SenHuang19 I see your point, I check that the `time` key is in the returned dictionary within the Julia examples [here](https://github.com/JavierArroyoBastida/project1-boptest/blob/13e71040c96afc723f4888dab118eaea438cb55b/examples/julia/twoday_p.jl#L45) and [here](https://github.com/JavierArroyoBastida/project1-boptest/blob/13e71040c96afc723f4888dab118eaea438cb55b/examples/julia/szvav_sup.jl#L45) because didn't find a way to directly check if return was a None. It may not be the most intuitive way to check that initialization was successful but is the best I could get with my Julia skills. Still I would avoid the overhead of a new variable definition in `testcase`, as @dhblum says, the controller can check if None is returned to determine if successful, there are probably several ways to do so.

dhblum:
I changed the julia code to check for ``nothing`` in https://github.com/ibpsa/project1-boptest/pull/225/commits/3f680ecbb13982f7ef44a741d7f04025efd198b1.  @JavierArroyoBastida Hopefully you don't mind me pushing to your fork.

Is that an ok fix @SenHuang19?  I'd like to move this along if so.

SenHuang19:
> I changed the julia code to check for `nothing` in [3f680ec](https://github.com/ibpsa/project1-boptest/commit/3f680ecbb13982f7ef44a741d7f04025efd198b1). @JavierArroyoBastida Hopefully you don't mind me pushing to your fork.
> 
> Is that an ok fix @SenHuang19? I'd like to move this along if so.

Yes, sounds good to me!

Issue #232, opened 2020-09-03
Use commits to set versions


JavierArroyoBastida:
@dhblum, some testcases use branch names instead of commits to set library versions of required dependencies. This practice is dangerous as branches are dynamic and can thus introduce new erros into the models. An example is what has happened in [this build](https://github.com/ibpsa/project1-boptest/runs/1062946080). I propose to always checkout commits of the dependency libraries and to perform manual (and thus controlled) updates when required.

dhblum:
I agree with this.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/233.

PR #233, opened 2020-09-03
Use commits to set versions

JavierArroyoBastida:
For https://github.com/ibpsa/project1-boptest/issues/232


dhblum:
@JavierArroyoBastida Thanks.  Are the commits added here the current master branches as of when you made the change (8 days ago)?

JavierArroyoBastida:
@dhblum yes, these remain the current master branches.

Issue #244, opened 2020-10-08
Different pricing scenarios in reported cost KPI


dhblum:
Currently, the ``/kpis`` API call returns the cost KPI for the constant pricing scenario.  There is no way for a user to get the cost for the dynamic and highly dynamic pricing scenarios.  This issue is to add an API for the user to select the pricing scenario for which cost should be calculated.  Or, to add to the existing KPI report all three costs.  I am in favor of the former, as eventually, I think we need to standardize how to setup a given test, and selecting a price scenario is part of this.

dhblum:
My proposal is to add:

Interaction|Request
-----------|--------
Receive current price scenario | GET ``price_scenario``
Set price scenario | PUT ``price_scenario`` with argument ``type=<'constant' or 'dynamic' or 'highly_dynamic'>``

JavierArroyoBastida:
I'm surprised that we've overlooked this till now since specifying price scenario is an important feature. The solution that you propose looks good to me, and also prefer that one over showing all three costs.

dhblum:
Implemented on https://github.com/ibpsa/project1-boptest/tree/issue244_setPriceScenario.

dhblum:
Another idea is to implement the following:

Interaction|Request
-----------|--------
Receive current scenario settings | GET ``scenario``
Set scenario settings | PUT ``scenario`` with argument json: ``test_settings = {'electricity_price' : <'constant' or 'dynamic' or 'highly_dynamic'>, 'time_period' : <'summer' or 'winter' or 'shoulder' or 'annual'>, 'weather_forecast_uncertainty' : <'none' or 'low' or 'high'>}`` 

In this way, we already have a mechanism for setting all scenario parameters.

JavierArroyoBastida:
That sounds good to me. I like having all scenario settings within the same dictionary. That will probably help with categorization of test case runs. Also agree with the scenarios that you propose. One think we may need to think though is whether the `annual` scenario needs to be initialized and how.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/250.

PR #250, opened 2020-10-29
Issue244 set price scenario

dhblum:
This is for #244.  It:

- Adds API for GET and PUT ``/price_scenario``
- Adds and updates unit tests
- Adds ``'constant'`` price scenario as default in ``config.py`` for test cases


dhblum:
@JavierArroyoBastida Before review, actually see this comments: https://github.com/ibpsa/project1-boptest/issues/244#issuecomment-721367335.

JavierArroyoBastida (approved):
I agree with the changes @dhblum, thanks for implementing this! A more complete implementation would indeed be including what is being described in [this comment](https://github.com/ibpsa/project1-boptest/issues/244#issuecomment-721367335). However, I'd go for a different pull request for that extension as this one is ready to merge.

dhblum:
Thanks @JavierArroyoBastida for the review.  I was thinking I could update this PR based on [the comment](https://github.com/ibpsa/project1-boptest/issues/244#issuecomment-721367335).  It should only be a matter of updating the API endpoint, but the backend stuff should be able to stay mostly the same.

dhblum:
@JavierArroyoBastida I've made updates based on that comment.

JavierArroyoBastida:
looking good to me as well. Thanks @dhblum!

Issue #264, opened 2020-12-11
Make KPI Calculator efficient


JavierArroyoBastida:
This is to reduce the memory usage related to the calculation of KPIs, following what has been discussed in https://github.com/ibpsa/project1-boptest/issues/240.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/265.

Issue #240, opened 2020-10-04
Memory usage


haraldwalnum:
When we have tested with longer simulation periods (e.g. 1 year), we have had some issues with the docker container crashing with OME errors (exit code 137).
When running docker on windows (through WSL) you need to allocate how much memory docker can use, so for now we have been able to solve this by increasing the memory allocation, but even with relatively simple emulators we need about 8GB of memory.
 
To investigate this further I did a test with the bestest_hydronic case. I run it for 180 days, with a control step of 1 hour. 
While running the simulation, I logged the memory allocation of the docker container. I have attached the output in a csv file and also a plot of the results.
The case is run three times. No control is applied, but the control step is still 1 hour, so the emulator is stopped and started again every hour (requests.post('{0}/advance'.format(url), data = {}).json())
After each case, results and KPIs are requested:
results = requests.get('{0}/results'.format(url)).json()
kpi = requests.get('{0}/kpi'.format(url)).json()
 
In the attached data, you can recognise a spike in the memory usage after each simulation. This corresponds to the get.result request. After that you can see 7 slightly smaller spikes, which corresponds to the get.kpi request.

![bestest_hydronic_memory](https://user-images.githubusercontent.com/53710940/95023736-ea363880-067e-11eb-9dbd-5ae4a2824cd0.png)
[bestest_hydronic_memory.txt](https://github.com/ibpsa/project1-boptest/files/5324037/bestest_hydronic_memory.txt)

 
My main concern is that this is still a relatively simple emulator, with few outputs, and that this will be a bigger issue with more complicated emulators.

This issue is probably linked with issue #124 and might be merged.

EttoreZ:
@haraldwalnum I developed a  custom testcase which is more complex and I also have memory issues for the specific run. Furthermore if I run several simulations in sequence (changing control strategy), without closing the docker container, the memory usage rises easily up to 16 GB.

dhblum:
I've done a little studying on this.  I repeated Harald's test with bestest_hydronic (180 days and step size of 1 hour) on Ubuntu 16.04.  I add a 60 second pause between calling ``/results`` (first spike at end of test) and calling ``/kpis`` (second spike at end of test).  I get:

![mem_baseline](https://user-images.githubusercontent.com/19480999/97629307-500aab80-1a04-11eb-95da-f28bf79a354e.png)

with KPIs:

```
KPI RESULTS 
-----------
ener_tot: 2844.72560261 kWh
[...]

PR #263, opened 2020-12-07
Add precommit and run whitespace cleanup

nllong:
This PR accomplishes the following

* Add a pre-commit config file
* Run recommit cleanup on whitespace and end-of-line fixes

There are a lot of files touched on this PR, but the changes are strictly cosmetic.

#262

PR #265, opened 2020-12-11
Issue264 make kpi cal efficient

JavierArroyoBastida:
This is for https://github.com/ibpsa/project1-boptest/issues/264: to reduce the memory usage related to the calculation of KPIs, following what has been discussed in https://github.com/ibpsa/project1-boptest/issues/240. Specifically:

1. It allows to call the KPI Calculator iteratively, which was technically possible before but computational time was increasing with simulation time as integration always happened since the starting simulation time. Now, storage variables are included so that new integrations can be calculated from the results of previous integrations. This is particularly useful for the implementation of reinforcement learning algorithms (see https://github.com/ibpsa/project1-boptest-gym), where KPIs may need to be calculated every sampling time to define a reward function. Notice that in https://github.com/ibpsa/project1-boptest/issues/240 we were considering two possibilities: the integration within Modelica models and the use of cumulative records. I've finally gone for the latter because this one does not intrude into the emulator models, hence it does not hamper their functionality. 

2. It filters the variables when using the `data_manager.get_data()` method in the `KPI_Calculator`, so that now only the needed variables are retrieved. Each of these variables need to be reindexed, so retriving only the needed ones is a notable improvement.  

Note that with this new approach the KPI Calculator needs to reset every time the scenario is changed, so that the integration in the new scenario is not based on the results from the previous scenario.


JavierArroyoBastida:
Unit tests are passing locally. For some reason Travis tests have not been triggered so far despite what has been discussed in https://github.com/ibpsa/project1-boptest/pull/263#issuecomment-743240878.

dhblum:
Thanks @JavierArroyoBastida.  I likely won't be able to review this until next week.  And I think the tests are running now.  We need to form a queue with all other ibpsa repos, including modelica-ibpsa.  So sometimes, we'll have to wait until its our turn for them to start.

JavierArroyoBastida:
Indeed, I see now they've already run. Thanks for clarifying @dhblum!

data/data_manager.py
```diff
@@ -237,7 +237,7 @@ def save_data_and_kpisjson(self, fmu_path):
         self.z_fmu.close()
         
     def get_data(self, horizon=24*3600, interval=None, index=None, 
-                 category=None, plot=False):
+                 variables=None, category=None, plot=False):
```
dhblum, line 240: Just a note that this may come in handy later too to make forecast data retrieval more efficient.

kpis/kpi_calculator.py
```diff
@@ -59,6 +59,97 @@ def __init__(self, testcase):
...
+        '''Initialize variables required for KPI calculation
+        
+        '''
+        # Initialize index since last integration
+        setattr(self, 'i_last_{}'.format(label), 0) 
+        # Dictionary to store energy usage by element
+        setattr(self, '{}_dict'.format(label), OrderedDict())
+        # Dictionary to store energy usage by source 
+        setattr(self, '{}_dict_by_source'.format(label), OrderedDict())
+
```
dhblum, line 80: Why the use of ``setattr`` instead of ``self.x = y``?

dhblum (changes requested):
Thanks a lot @JavierArroyoBastida for this implementation.  In general it looks good and seems to do the job.  Just one question I had, see the inline comment.  Before merging, still need to:

- merge latest master and resolve conflicts
- pass pre-commit checks

JavierArroyoBastida:
Thanks a lot for your review @dhblum! I've just merged master and resolved conflicts, waiting now for tests to pass.

kpis/kpi_calculator.py
JavierArroyoBastida: This is just to enable an attribute with a dynamic name that can be defined from a string using `format()`

dhblum: Ah ok I see now.  Ok sounds good, thanks.

dhblum:
Thanks @JavierArroyoBastida for resolving the conflicts and merging latest master.  I ran pre-commit checks and there were some whitespace issues so committed a few additional changes.  Will merge after tests pass.

dhblum (approved):

Issue #267, opened 2020-12-17
Add data slice arguments to /results


dhblum:
Currently, ``/results`` retrieves data for all input and outputs variables for all time since the beginning of the simulation.  As pointed out in https://github.com/ibpsa/project1-boptest/issues/240, this is very inefficient in terms of memory usage and computational time if the user is only looking for data on a subset of variables or specific point in time.  It will get worse with larger models.  In addition, typical BAS databases require specification of the point name and time period for data collection.

Therefore, this issue is to revise the ``/results`` api.  I propose adding three arguments to the api call:

- point_name: <str> the name of point to collect data for
- start_time: <int> the beginning time of the data slice in seconds
- final_time: <int> the ending time of the data slice in seconds

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/268.

PR #268, opened 2020-12-22
Issue267 results api update

dhblum:
This is for #267.  It does the following:

- Change ``/results`` to PUT request with arguments ``point_name``, ``start_time``, and ``final_time``
- Updates interfaces for tests and examples
- Adds new tests for obtaining result data with partial time slices, with start and final times both before and after the start and end time of a simulation.
- Updates readme


dhblum:
@SenHuang19 @JavierArroyoBastida Can you please review this?  It is my proposed API change for getting results one variable at a time and for specific time periods.

dhblum:
The one ci build failed because of the julia image not being able to be pulled.

README.md
```diff
@@ -55,7 +55,7 @@ Example RESTful interaction:
 | Set communication step in seconds                                     |  PUT ``step`` with argument ``step=<value>``              |
 | Receive sensor signal names (y) and metadata                          |  GET ``measurements``                                     |
 | Receive control signals names (u) and metadata                        |  GET ``inputs``                                           |
-| Receive test result data                                              |  GET ``results``                                          |
+| Receive test result data for the given point name between the start and final time in seconds |  PUT ``results`` with arguments ``point_name=<string>``, ``start_time=<value>``, ``final_time=<value>``|
```
JavierArroyoBastida, line 58: Is there a reason to use the "point" keyword? the keyword "variable" feels more intuitive to me. I see a point being a set of variable values at a certain time whereas here we're talking about time series. Does "point" belong to the BMS jargon? in that case I'd have nothing to say. If not, I'd suggest to change it to "variable".

testing/utilities.py
```diff
@@ -272,13 +272,20 @@ def create_test_points(self, s,n=500):
 
         return s_test
 
-    def results_to_df(self, results):
+    def results_to_df(self, points, start_time, final_time, url='localhost:5000'):
```
JavierArroyoBastida, line 275: `url` argument is not being used here. I'd suggest to remove it.

testing/utilities.py
```diff
@@ -523,3 +548,35 @@ def test_set_scenario(self):
...
+
+    def test_partial_results_inner(self):
+        '''Test getting results for start time after and final time before.
+
+        '''
+
+        requests.put('{0}/initialize'.format(self.url), data={'start_time':0, 'warmup_period':0})
+        requests.put('{0}/step'.format(self.url), data={'step':self.step_ref})
+        measurements = requests.get('{0}/measurements'.format(self.url)).json()
+        y = requests.post('{0}/advance'.format(self.url), data=dict()).json()
```
JavierArroyoBastida, line 560: Variable `y` is not used after assignment. I'd suggest to remove or to change by `_`

testing/utilities.py
```diff
@@ -523,3 +548,35 @@ def test_set_scenario(self):
...
+
+    def test_partial_results_outer(self):
+        '''Test getting results for start time before and final time after.
+
+        '''
+
+        requests.put('{0}/initialize'.format(self.url), data={'start_time':0, 'warmup_period':0})
+        requests.put('{0}/step'.format(self.url), data={'step':self.step_ref})
+        measurements = requests.get('{0}/measurements'.format(self.url)).json()
+        y = requests.post('{0}/advance'.format(self.url), data=dict()).json()
```
JavierArroyoBastida, line 576: Variable `y` is not used after assignment. I'd suggest to remove or to change by `_`

JavierArroyoBastida (changes requested):
Thanks a lot @dhblum for implementing this. Looks good in general. I've just added a few comments/suggestions that you can find inlined.

README.md
dhblum: A "point" is BMS jargon for input/output variable.  Part of a BMS/Controls design is a "points list," which lists all of the sensor and control "points."

testing/utilities.py
dhblum: Addressed in https://github.com/ibpsa/project1-boptest/pull/268/commits/3d4a58056dc44ff1340ed5dd256001f9b9e5b6b7.  Actually implemented to use ``url`` instead of not since I think it makes the code easier to follow.  Also ``get_all_points`` uses the ``url`` argument.

testing/utilities.py
dhblum: Addressed in https://github.com/ibpsa/project1-boptest/pull/268/commits/ba2364682a3d59e684cfb42916f1e23050b8c4c0.

testing/utilities.py
dhblum: Addressed in https://github.com/ibpsa/project1-boptest/pull/268/commits/ba2364682a3d59e684cfb42916f1e23050b8c4c0.

README.md
JavierArroyoBastida: Great, let's keep it like this then.

JavierArroyoBastida (approved):

README.md
dhblum: Sounds good.  I added "point" to the readme description of /measurements and /inputs in  https://github.com/ibpsa/project1-boptest/pull/268/commits/d1c0eaa5353713ac3da28bf595b2259bd2aeb538.

dhblum:
@SenHuang19 I've updated the javascript example with the new results api example proposed in this PR.  It only gets results for zone air temperature.  I couldn't figure out the required for loops in javascript, but it is ok by me for unit tests to just check this one variable.  Can you please check this and rest of PR is ok with you when you get a chance?  Thanks!

SenHuang19:
@dhblum it looks good to me. We can improve it when updating the example files. Should we create a new issue for updating the example files?

dhblum:
@SenHuang19 Yes I'm ok with opening a new issue if there are additional updates to the example file you want to do.  Can you identify what they are in that issue?

dhblum:
Closes #267

PR #271, opened 2021-01-10
Issue112 browser access additional modifications

SenHuang19:


dhblum:
@SenHuang19 Is it necessary to include the geckodriver file in the repository?  I think I'd rather leave it as a dependency.

SenHuang19:
@dhblum fixed the issue associated with geckodriver

dhblum:
Awesome, thanks @SenHuang19!  Tests pass too, so I will merge.

Issue #274, opened 2021-01-21
TestCase: Bestest Hydronic pump control


haraldwalnum:
@dhblum @JavierArroyoBastida . As discussed yesterday I had some issues in controlling the pump on/off in the besttest hydronic testcase. I have tried to reproduce the issue in a simple test setup. I run the emulator with 3600s control step, turning the pump on/off for each step. The results are show below. One can see that ovePump_u and reaPump_y follows the control signal, but reaPPum_y consumes power more "randomly". This is also reflected by the gas consumed by the heater (reaQHea_y). 

![pumpTest](https://user-images.githubusercontent.com/53710940/105298985-c3165f00-5bba-11eb-8283-5acf5bfe579b.png)

I have also attached the code I used to run the test.
[pumpTest.txt](https://github.com/ibpsa/project1-boptest/files/5847349/pumpTest.txt)

JavierArroyoBastida:
Thanks @haraldwalnum for sharing the issue and script, I could easily reproduce the problem. I've been doing some testing on this and it's indeed an awkward behavior. I'm afraid I've not been able to realize the source of the problem, but have two comments on this:
1. Not sure if we're checking out the right commit of IDEAS for this testcase: we're using now `a91fd24d97a9dff737501a94cb6387530d618f6c` but I couldn't find that commit in the IDEAS library... not sure where it's coming from, maybe it's just me but just couldn't find it. I think we should check out: `f90badebc912251714a4ca93c86739cca34ea044`. Still, nothing has changed when checking out the new commit. 
2. I have extended the model to have a "debug model" that reads more variables which might help us to figure out what's going on. Basically, I've added read blocks for the integer that sets the pump stage, the pressure difference in the pump and the normalized pump speed. This debug model looks as follows:

![image](https://user-images.githubusercontent.com/22677958/105339916-8314a380-5bdd-11eb-9bb0-f09aa6ce6634.png)

See results here (also with indoor temperature):

![image](https://user-images.githubusercontent.com/22677958/105340692-83616e80-5bde-11eb-8e6e-d260742d23d2.png)

What is interesting from here is that the integer variable that sets the pump stage is already corrupted, which indicates that the problem comes from the `realToInteger`. The doc of this model:

<img src="https://user-images.githubusercontent.com/22677958/105343463-f28c9200-5be1-11eb-9160-a924bc852f53.png" width="600">

[...]

PR #284, opened 2021-02-05
Fix for event update at start of simulation step

dhblum:
This PR is for https://github.com/ibpsa/project1-boptest/issues/274, though implements a generic fix for event handling at the start of a simulation step.  In particular, it:

- Adds a test for improper event handling using the ``bestest_hydronic_heat_pump`` test case as described in #274 
- Adds a line to ``Dockerfile`` that writes code in ``pyfmi`` to fix event handling at start of simulation step
- Adds correct test reference results, verified by comparison to results generated by Dymola with Dassl solver tol 1e-6 and use of ``Modelica.Blocks.Sources.Pulse`` for pump signal injection
- Updates reference results for testcase2, which had very small numeric differences


JavierArroyoBastida (approved):
Looking good to me. Thanks @dhblum for getting this implemented.

Issue #270, opened 2021-01-07
Specifying scenario time period


dhblum:
This issue is to enable the ``'time_period'`` parameter proposed in #244 in the comment https://github.com/ibpsa/project1-boptest/issues/244#issuecomment-721367335 for some initial test scenario definitions.

First, we must specify the time periods for each emulator.  I propose:

- We define test scenarios that last for two weeks, with a one week warmup-period utilizing baseline control.  The time period argument is defined by the middle day of the two weeks such that:

1. "cooling_peak" = day with the max 15-minute system cooling load in the year.
2. "cooling_typical" = day with the max 15-minute system cooling load that is closest from below to the median of all 15-minute maximum cooling loads of all days in the year.
3. "heating_peak" = day with the max 15-minute system heating load in the year.
4. "heating_typical" = day with the max 15-minute system heating load that is closest from below to the median of all 15-minute maximum heating loads of all days in the year.

- Systems that do not have cooling will only have the heating scenarios available.
- We'd need to only consider days 8 through 357 in the above determination if we want to avoid test scenarios that go through end/start of year.  

Then, for implementing in code.  I propose:

- We use the ``config.py`` file to define the days outlined above.
- ``testcase.py`` utilizes this information to invoke the ``initialize()`` function appropriately and define a flag that can be used to stop a simulation upon reaching the end of the time period.

Feedback would be great!

JavierArroyoBastida:
This approach sounds reasonable to me. We may want to develop a script that upon being called runs the baseline controller and identifies the test scenarions as you propose. 
One doubt I have is: why in the `_typical` scenarios does the load need to be the closest from *below* (and not above) the median for both: heating and cooling?
Also for the implementation I think what you suggest makes sense but, is it the intention then to remove the `start_time` and `warmup_time` arguments from the `initialize` API call? I would let these be optionally used: if not specified, then the default is that the testcase parses the period from the scenario. If used, then the scenario is disregarded and the value sent is used. The reason I'm pointing this out is that there should be some flexibility to use the testcase for training models/algorithms out of the testing period.

dhblum:
Thanks @JavierArroyoBastida for the feedback.  

To your first point, I agree and already started.
To your second point, from above seems just as valid I suppose.  Would need to pick one.
[...]

PR #285, opened 2021-02-05
Issue270 scenario time

JavierArroyoBastida:
@dhblum this extends your branch for finding the days to define the scenarios. A few notes:
1. Functionality for fining days remains exactly the same as your original script. 
2. As this functionality is common to all emulators I thought would be a good idea to move elsewhere, e.g. the `data` folder. 
3. The main extension is the possibility to retrieve the data using a yearly simulation of a deployed testcase. 
4. Still need to figure out how to find the days for periods when heating and cooling are interchangeably demanded. 
5. Still need to run the pre-commit checks.

EDIT: pre-commit checks are passed now in https://github.com/ibpsa/project1-boptest/pull/285/commits/6ac04de64d30b714603b2f23bd4d19ca102f0630.


testcases/bestest_air/models/find_days.py
```diff
@@ -6,28 +6,10 @@
...
 """
 
-import pandas as pd
+from data.find_days import find_days
+import json
 
-csv_filepath = 'TestCase_Ideal.csv' # csv file containing model results (with Time as index column in seconds)
-heat = 'fcu.powHeaThe.y' # Column for heating load [W]
-cool = 'fcu.powCooThe.y' # Column for cooling load [W]
+days = find_days(heat='fcu_reaPCoo_y', cool='fcu_reaPHea_y', data='simulate')
```
dhblum, line 12: I think we would rather want the thermal load to be used instead of electrical or gas power, so as not to take COP changes (if they exist) into account.  The variables are as before:

heat = 'fcu.powHeaThe.y' # Column for heating load [W]	
cool = 'fcu.powCooThe.y' # Column for cooling load [W]

testcases/bestest_hydronic_heat_pump/models/find_days.py
```diff
@@ -0,0 +1,17 @@
...
+
+Finds the peak and typical heat days for the bestest_hydronic_heatpump
+case. This case needs to be deployed.
+
+'''
+
+from data.find_days import find_days
+import json
+
+days = find_days(heat='reaPHeaPum_y', cool=None, data='simulate')
```
dhblum, line 14: I think we would rather want the thermal load to be used instead of electrical or gas power, so as not to take COP changes (if they exist) into account.

testing/test_data.py
```diff
@@ -341,5 +342,35 @@ def setUp(self):
...
+                            'data', 'find_days', 'days_ref.json')
+
+
+    def test_find_days(self):
+        '''The test uses one month simulation data as obtained from
+        the bestest_air case.
+
+        '''
+
+        days = find_days(heat='fcu_reaPCoo_y', cool='fcu_reaPHea_y',
```
dhblum, line 370: I suggest testing with using thermal load instead of electrical or gas power.

dhblum (changes requested):
Thanks a lot @JavierArroyoBastida for extending this.  Generally looks good, but a couple of comments on the use of thermal load instead of electrical/gas power.

testing/test_data.py
JavierArroyoBastida: Addressed in https://github.com/ibpsa/project1-boptest/pull/285/commits/a0e7d9e84d91714e6b561aa9682424ab11f83fb5.

testcases/bestest_hydronic_heat_pump/models/find_days.py
JavierArroyoBastida: Addressed in https://github.com/ibpsa/project1-boptest/pull/285/commits/78056613cfdda5b15e0204c039d2fd98b6a23c92.

testcases/bestest_air/models/find_days.py
JavierArroyoBastida: Addressed in https://github.com/ibpsa/project1-boptest/pull/285/commits/d586aa78278bef0757a972ad538fbfa5ccb3f979.

JavierArroyoBastida:
Yeah good point, thanks for the observation! the new commits should address that.

dhblum (approved):
Changes look good, thanks!

dhblum:
@JavierArroyoBastida Should I merge this?  It is any case not to the master.  But there is still 4. of your list to be completed (good suggestion in https://github.com/ibpsa/project1-boptest/issues/270#issuecomment-774719263) and also an additional item not listed to connect the ``/scenario`` API with your proposed ``days.json`` to actually perform the appropriate initialization.  I'm ok merging this PR as a "part 1" or something.

JavierArroyoBastida:
Yes, I'd suggest merging this one already and keep working from here. I'll send new a PR (with something like "part 2") to this same branch if I find the time to implement what @haraldwalnum proposed in https://github.com/ibpsa/project1-boptest/issues/270#issuecomment-774719263.

dhblum:
Sounds good to me.  Let me know if you do find the time, and I'll do the same, so as not to duplicate.

PR #281, opened 2021-02-02
Sync master and boptest-service

nllong:
Hi @dhblum -- Can you merge master into boptest-service, it is a clean merge? Thanks!

Issue #278, opened 2021-01-26
Configuration error


dhblum:
When running ``examples/python/twoday_p_service_api.py`` an error is thrown:

```
worker_1  | Traceback (most recent call last):
worker_1  |   File "/boptest/step_sim/step_fmu.py", line 384, in <module>
worker_1  |     runFMUSite = RunFMUSite(site_ref=site_ref, real_time_flag=real_time_flag, time_scale=time_scale, startTime=startTime, endTime=endTime, externalClock=externalClock)
worker_1  |   File "/boptest/step_sim/step_fmu.py", line 111, in __init__
worker_1  |     self.tc = boptest.lib.testcase.TestCase()
worker_1  |   File "/boptest/lib/testcase.py", line 58, in __init__
worker_1  |     self.set_scenario(con['scenario'])
worker_1  | KeyError: 'scenario'
```

This is because the configuration of the test case needs to be updated in the worker to include the scenario.  Currently, the configuration is set by ``step_fmu.py`` lines 97-105, but needs to be updated to align with at least ``testcases/testcase1/config.py``.  

A more comprehensive fix may be looking at packaging this configuration in the resource directory of the test case FMU as a way to load it dynamically.

Issue #276, opened 2021-01-23
Update /results api


dhblum:
Following #267, this issue is to update usage of the /results api.  This could be done first by:

- [ ] 1. Replicating existing functionality with a for-loop through all point names.  

But an improvement would be to have the web service deliver the specified result data on-demand, during a simulation.  One option for this is to pass data from the worker to the web service upon being called to do so via the redis database, in a similar pattern as is done for /advance.

PR #288, opened 2021-02-10
service-master sync

dhblum:
This is in place of https://github.com/ibpsa/project1-boptest/pull/281 with fixes for #276 and #278 to replicate boptest-service functionality from before master api changes.

Note that the fix for #276 replicates the functionality of boptest-service from before the new /results api (update all results at end of simulation).  It does not update boptest-service to reflect the new capabilities of the /results api (make it possible to retrieve results for one variable for specific time period).

Note that the fix for #278 sets the prices scenario to ``"constant"``, but does not add the fully featured ``/scenario`` api to boptest-service. 

@kbenne With this PR, running ``examples/python/twoday_p_service_api.py`` works until reporting of KPIs.  The KPIs return as all 0s.  I spent a bit of time trying to figure out why but couldn't.  Even Line 230 of ``step_fmu.py`` returns KPIs with 0s.  Is the testcase (``self.tc``) somehow re-initialized or purged of it's history between when the simulation ends and ``cleanup()`` begins?


dhblum:
@kbenne I think this PR can be closed with all of the recent updates to the ``boptest-service`` branch.  But please confirm, and close if indeed this is true.

kbenne:
yep. service is current with master. I'll go ahead and close.

Issue #280, opened 2021-01-29
Access results for initialization time


dhblum:
An MPC or other controller may need access to the warmup data from the /initialize API call one would expect from using /results to do state estimation or other calculation at the start of a test.  This issue is to make that happen.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/290.

PR #290, opened 2021-02-18
Issue280 results available for initialization

dhblum:
This is for #280.


dhblum:
Remaining to do on this:
- Add initial time to KPI calculation instead of always using data that starts from 0 seconds, since now some of that can be initialization data.

dhblum:
@JavierArroyoBastida Can you review this when you get a chance?  It enables storage of initialization data as in #280, but there was some work to be done on the kpi calculator so as not to use that data for kpi calculation.

testing/references/kpis/cost_dict_MultiZone.csv
```diff
@@ -0,0 +1,3 @@
+keys,value
```
JavierArroyoBastida, line 1: Is it possible that this reference file is coming from previous checkouts of the repo? The "cost" references should start by either of the following labels: `cost_constant`, `cost_dynamic` or `cost_highly_dynamic`. If that's the case this file must be removed.

testing/references/kpis/cost_dict_SingleZone.csv
```diff
@@ -0,0 +1,5 @@
+keys,value
```
JavierArroyoBastida, line 1: Same comment as in `cost_dict_MultiZone.csv`.

testing/references/kpis/cost_tot_MultiZone.csv
```diff
@@ -0,0 +1,2 @@
+keys,value
```
JavierArroyoBastida, line 1: Same comment as in `cost_dict_MultiZone.csv`.

testing/references/kpis/cost_tot_SingleZone.csv
```diff
@@ -0,0 +1,2 @@
+keys,value
```
JavierArroyoBastida, line 1: Same comment as in `cost_dict_MultiZone.csv`.

testing/test_kpis.py
```diff
@@ -93,15 +93,15 @@ def test_get_cost(self):
         # Check results
         self._perform_test(self.case.cost_tot, self.case.cost_dict, 'cost_constant')
 
-        # Reset KPI Calculator
-        self.cal.initialize()
+        # Reset testcase and kpi calculator
+        self._initialize_testcase(self.testcase_name, self.ref_data_filepath)
```
JavierArroyoBastida, line 97: I'm struggling a bit with this one. Is it really needed to initialize the testcase as well? actually, is it correct? does not the simulation data get lost when re-initializing the case? Apparently not since tests are passing, but that's surprising to me.

JavierArroyoBastida: I see now, simulation data is not lost because it is reloaded in `_initialize_testcase()`. Still, wouldn't we get the same result by just resetting the KPI Caluclator with `self.cal.initialize()`? I like more that approach since that also checks that the KPI Calculator can re-calculate KPIs without having to re-run the simulation even after changing the scenario.

JavierArroyoBastida (changes requested):
Looking good in general. Just added a few suggestions and comments.

kpis/kpi_calculator.py
```diff
@@ -532,6 +532,44 @@ def get_computational_time_ratio(self, plot=False):
 
         return time_rat
 
+    def _set_last_index(self,label):
```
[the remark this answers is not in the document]
dhblum: Actually, I think label is ok here because it actually does not reference a test case point name, rather the suffix used for the variable that keeps track of the last index in the kpi calculator.  The description was incorrect (not updated from an earlier version of the function I had tried out).  I've updated the description in this commit: https://github.com/ibpsa/project1-boptest/pull/290/commits/e38a5e62353d82e30f2b27b3b08bd1a601a715fc.

kpis/kpi_calculator.py
```diff
@@ -532,6 +532,44 @@ def get_computational_time_ratio(self, plot=False):
...
+        '''Set last index for kpi calcualtion.
+
+        Parameters
+        ----------
+        label: str
+            Name of point to get data for from case.y_store
```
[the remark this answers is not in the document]
dhblum: The description was incorrect in the first place (not updated from an earlier version of the function I had tried out). I've updated the description in this commit: e38a5e6.

kpis/kpi_calculator.py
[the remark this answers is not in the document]
dhblum: To your general point though, I agree.  I've changed the name of the variable in ``_get_data_from_last_index`` accordingly in https://github.com/ibpsa/project1-boptest/pull/290/commits/7d63d36db8b5fd8ba53d7345f426827197638b96.

kpis/kpi_calculator.py
```diff
@@ -532,6 +532,44 @@ def get_computational_time_ratio(self, plot=False):
...
+        setattr(self, 'i_last_{}'.format(label),i)
+
+    def _get_data_from_last_index(self,key,t):
+        '''Get data from last index indicated by t.
+
+        Parameters
+        ----------
+        key: str
+            Name of point to get data for from case.y_store
+        t: num
```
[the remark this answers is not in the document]
dhblum: Again, this description was not updated from an earlier version of the function.  I'm sorry and thanks for catching.  ``t`` here should actually be an integer, since it represents an index.  As such, I suggest changing it to ``i`` and updating the description accordingly in https://github.com/ibpsa/project1-boptest/pull/290/commits/7d63d36db8b5fd8ba53d7345f426827197638b96.

kpis/kpi_calculator.py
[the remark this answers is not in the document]
dhblum: Cool, thanks!  Makes it easier to make changes to the method, which is then utilized throughout the module.

kpis/kpi_calculator.py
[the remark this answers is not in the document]
JavierArroyoBastida: Looking more clear now, thanks!

testcase.py
```diff
@@ -42,14 +42,16 @@ def __init__(self):
...
         self.forecaster = Forecaster(testcase=self)
-        # Instantiate a KPI calculator for the test case
-        self.cal = KPI_Calculator(testcase=self)
         # Get available control inputs and outputs
         self.input_names = self.fmu.get_model_variables(causality = 2).keys()
         self.output_names = self.fmu.get_model_variables(causality = 3).keys()
         # Get input and output meta-data
         self.inputs_metadata = self._get_var_metadata(self.fmu, self.input_names, inputs=True)
         self.outputs_metadata = self._get_var_metadata(self.fmu, self.output_names)
+        # Initialize simulation data arrays
```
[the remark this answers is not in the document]
dhblum: The initialization of the kpi_calculator depends on ``testcase.y_store`` from the new method ``_set_last_index``.  I suppose the order could be arbitrary if there is a check in ``_set_last_index`` for the existence of ``testcase.y_store`` in the first place.

testcase.py
```diff
@@ -264,10 +269,11 @@ def initialize(self, start_time, warmup_period):
         # Process result
         if res is not None:
             # Get result
-            self.__get_results(res, store=False)
+            self.__get_results(res, store=True, store_initial=True)
```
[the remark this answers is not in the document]
dhblum: Done in https://github.com/ibpsa/project1-boptest/pull/290/commits/9570b36ff1ee020c63a7c04737747092fa55ef6d.

testing/test_kpis.py
```diff
@@ -207,29 +207,25 @@ def _perform_test(self, tot, dictionary, label):
...
-
-    def setUp(self):
-        '''Setup for each test.
+    def _initialize_testcase(self, testcase, ref_data_filepath):
+        '''Initialize the test case.
 
         '''
-
-        self.name = 'SingleZone'
         # Change directory to testcase 2
```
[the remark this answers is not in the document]
dhblum: Accepted your suggestion in https://github.com/ibpsa/project1-boptest/pull/290/commits/950e509548506cb2d524afdc59fd68048ebb4f5e.  Pretty cool feature I didn't know about.

testing/references/kpis/cost_dict_MultiZone.csv
dhblum: This reference was new when I ran the tests locally.  I think it is from ``test_iterative_call``.

testing/references/kpis/cost_dict_SingleZone.csv
dhblum: This reference was new when I ran the tests locally.  I think it is from ``test_iterative_call``.

testing/references/kpis/cost_tot_MultiZone.csv
dhblum: This reference was new when I ran the tests locally.  I think it is from ``test_iterative_call``.

testing/references/kpis/cost_tot_SingleZone.csv
dhblum: This reference was new when I ran the tests locally.  I think it is from ``test_iterative_call``.

testing/test_kpis.py
dhblum: No you do not get the same result.  Using ``self.cal.initialize()`` resets the stored kpi value to 0 using ``initialize_kpi_vars`` and resets the last index using ``_set_last_index``.  The last index will point to the end of the already existing ``y_store``.  So, when the kpi calculator is invoked the next time, you will get the kpi to be equal to 0 + 0 = 0.  The first 0 being from the reset of the kpi value itself, and the second 0 being from the fact that the the last index is already at the end of the ``y_store``. 

So, in the current implementation, you cannot change the scenario and re-calculate the KPIs from the beginning of the simulation (while also excluding the initialization period) with the new scenario.  To do so, we'd have to implement a way for ``_set_last_index`` to save the index it obtained when the test case itself was initialized, and then know to go back to that index if only the kpi calculator is re-initialized (and not also the test case).  I guess the question is how needed is that functionality.  I guess the one use case I think of is if you want to re-calculate the cost KPI for an alternative scenario then the one you ran your controller for (which may or may not have used the initial price scenario for control).

JavierArroyoBastida: I see now, thanks for the explanation! I need to sleep on it a bit. Let me come back to it next Monday.

testing/references/kpis/cost_dict_MultiZone.csv
JavierArroyoBastida: True, forgot about that one.

dhblum:
I'm resolving the conflicts in this one due to changes from PRs https://github.com/ibpsa/project1-boptest/pull/291 and https://github.com/ibpsa/project1-boptest/pull/296.

JavierArroyoBastida:
great thanks!

dhblum:
@JavierArroyoBastida I think this is ready to merge once tests pass.  With all the updates to bestest_hydronic_heat_pump in recent merges to master, kept having to resolve conflicts with updated reference results.  But double checked everything and should be good though.  So if you approve after a final review, can you do so using the review function?

JavierArroyoBastida (approved):

JavierArroyoBastida:
I've done a final review and think this can be merged.

Issue #279, opened 2021-01-27
Eliminate duplicate code in example controllers


SenHuang19:
This issue is to create generic main interface functions to reduce duplicate code in example controllers

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/376.

PR #292, opened 2021-02-23
Issue277 example controllers

SenHuang19:
Creating a generic python function for handling the data exchange between emulators and controllers


dhblum:
This for #279.

examples/python/interface.py
```diff
@@ -0,0 +1,172 @@
...
+
+def control_test(config):
+    '''Run test case.
+
+    Parameters
+    ----------
+    plot : bool, optional
+        True to plot timeseries results.
+        Default is False.
+    config : string, dict
```
dhblum, line 28: I think it'd be better to remove the ``config`` construct and instead just have keyword parameters like ``step``, ``length``, etc. with proper descriptions of each in the docstring.

examples/python/interface.py
```diff
@@ -0,0 +1,172 @@
...
+    if prediction_config is not None:
+        predictions_store = pd.DataFrame(columns=prediction_config)
+    else:  
+        predictions_store = None    
+    # Simulation Loop
+    for i in range(int(length/step)):
+        # Advance simulation
+        y = requests.post('{0}/advance'.format(url), data=u).json()
+        # Compute next control signal                
+        if prediction_config is not None:
```
dhblum, line 118: I understand this is to accommodate the original implementation of testcase3.py.  However, I'm wondering if it is better to abstract the use of "predictions" to within each controller, rather than in this interface, so that this interface can be read/interpreted more easily.  Then, simply provide the ``/forecast`` result to the imported controller (e.g. as in line 124) and let the imported controller parse the forecast as needed.  Basically, move lines 120-123 to within the controller.  Then, maybe ``prediction_config`` just needs to be a True/False, which is easier to interpret when reading the interface code.  And I think it'd be ok to remove the storing and plotting (in testcase3.py) of forecasted set point values, otherwise, store within the controller and add a method called "get_setpoints" or something to be used later in the plotting.

dhblum (changes requested):
See in line comments.

examples/python/interface.py
SenHuang19: I am not sure if it is extensible. How about leaving those details to the configuration files?

examples/python/interface.py
SenHuang19: Should we make the "interface" a class and extend this class with addition attributes for testcase3?

examples/python/interface.py
dhblum: I'm sorry I'm not sure what you mean by extensible.  My suggestion is to have:
```
def control_test(control_module, length, step, customized_kpi_config=None, prediction_config=None):
```
or if not, then to define each expected keyword of the argument ``config`` in the docstring, so it is clear what each configuration parameter represents.

I also suggest to change ``prediction_config`` to ``forecast_config`` so as to be consistent with the API call it refers to so as to improve readability.

examples/python/interface.py
dhblum: I'm not sure this is necessary, I'd be worried it may overcomplicate the readability of the code.  My suggestion relates to putting more of the details related to the parsing of the return ``forecast`` into the individual controller itself (since, abstractly, different controllers may utilize the predictions differently) rather than the generic interface, so as to make the reading of the generic interface code simpler.  That is, a reader sees whether a controller uses a forecast or not and how it may retrieve it.  If they want to know more about actually how this specific controller uses the forecast, they can go into the controller itself to view more.  But it is clear that the way the controller uses the forecast is specific to that controller, and not generic.

dhblum:
@SenHuang19 Is this ready for another review or still being worked on?

examples/python/interface.py
SenHuang19: As suggested, we changed the input arguments of control_test

examples/python/interface.py
SenHuang19: As suggested, we put the forcast parsing into individual controller

SenHuang19:
> @SenHuang19 Is this ready for another review or still being worked on?

It is ready for another review

examples/python/controllers/pidTwoZones.py
```diff
@@ -74,11 +79,41 @@ def initialize():
...
+        'oveActNor_u': 0,
+        'oveActNor_activate': 1,
+        'oveActSou_u': 0,
+        'oveActSou_activate': 1
+    }
 
     return u
+
+
+def update_predictions(prediction_config, forecast):
```
dhblum, line 94: I suggest using "forecast" terminology instead of "prediction" terminology to maintain consistency with the BOPTEST forecast api and so it is easier for a reader to associate the connection with it.

examples/python/testcase1.py
```diff
@@ -9,22 +9,14 @@
 
 # GENERAL PACKAGE IMPORT
 # ----------------------
-import requests
-import numpy as np
-from examples.python.custom_kpi import custom_kpi_calculator as kpicalculation
-import json,collections
-import pandas as pd
+import sys
```
dhblum, line 12: The doc string description of this module ``testcase1.py`` is the same as ``interface.py`` and needs to be corrected.

examples/python/testcase2.py
```diff
@@ -9,22 +9,14 @@
 
 # GENERAL PACKAGE IMPORT
 # ----------------------
-import requests
-import time
-from examples.python.custom_kpi import custom_kpi_calculator as kpicalculation
-import json,collections
-import pandas as pd
+import sys
```
dhblum, line 12: Similar comment as before on ``testcase1.py`` doc string.  Check doc string description of this ``testcase2.py`` module, since it is the same as the ``interface.py`` module.

examples/python/interface.py
```diff
@@ -0,0 +1,158 @@
...
+    u = control.initialize()
+    # Store prediction if any
+    # Simulation Loop
+    forecast = None
+    for i in range(int(length/step)):
+        # Advance simulation
+        y = requests.post('{0}/advance'.format(url), data=u).json()
+        if control.use_forecast:
+            forecast_data = requests.get('{0}/forecast'.format(url)).json()
+            # Compute next control signal
```
dhblum, line 111: Incorrect comment for next line of code.

examples/python/controllers/controller.py
```diff
@@ -0,0 +1,58 @@
+import sys
+import importlib
+import pandas as pd
+
+
+class Control(object):
```
dhblum, line 6: Suggest renaming to "Controller" since it is a "Controller" object.

dhblum (changes requested):
@SenHuang19 Thank you very much for the updates, they look good for the most part.  Please see in line comments for a few further suggestions.  Also, I think ``examples/python`` would benefit from have a ``README.md`` describing the contents and relationships of ``interface.py``, ``/controllers``, and ``testcase1/2/3.py``.

examples/python/controllers/pidTwoZones.py
SenHuang19: As suggested, I changed it into "update_forecast"

examples/python/testcase1.py
SenHuang19: I updated the doc string

examples/python/testcase2.py
SenHuang19: I updated the doc string

examples/python/controllers/controller.py
SenHuang19: I changed it into "Controller"

examples/python/interface.py
SenHuang19: I updated the comment

examples/python/testcase2.py
dhblum: In testcase2 doc string, please change "feedback controller" to "supervisory controller".

examples/python/testcase1.py
dhblum: Great thanks!

examples/python/controllers/pidTwoZones.py
dhblum: My suggestion was to change all references to "prediction" to "forecast," including in the controller modules (e.g. predictions_store, testcase1.prediction_store, Controller.prediction, etc.)

examples/python/interface.py
dhblum: Thank you!

examples/python/controllers/controller.py
dhblum: Thank you!

dhblum:
I will merge to ibpsa branch and suggest some edits.

PR #297, opened 2021-03-01
Issue280b res initial

JavierArroyoBastida:
This is to address https://github.com/ibpsa/project1-boptest/pull/290#discussion_r583731409.


dhblum:
1. The KPI calculator is not re-initialized if the /scenario API is called.  This would have to be added to make the functionality available to a user.
2. The new testcase.initial_time should be assigned in testcase.__init__() too.

JavierArroyoBastida:
1. The KPI calculator is already being re-initialized when changing the scenario here: https://github.com/ibpsa/project1-boptest/blob/c5e115d176175e4b22617c3bd559a8fdc52e446b/restapi.py#L176 Is that what you're referring to or is it something else?
2. Addressed in https://github.com/ibpsa/project1-boptest/pull/297/commits/3772abc5bf267992e6b4676886c6b8c5580daab9.

dhblum:
Ah, you're right about 1.  I missed it.  Can we move that re-initialization to ``testcase.set_scenario()``?  I think best would be to keep the API interacting only with the ``testcase.py`` API, and let ``testcase.py`` manage interactions with other elements as needed.

JavierArroyoBastida:
That actually makes a lot of sense. Changed here: https://github.com/ibpsa/project1-boptest/pull/297/commits/6249fd0dfa8693e864586d3af92570ac86393f4b.

Issue #299, opened 2021-03-02
Testcase: bestest_hydronic_heat_pump setpoint control


haraldwalnum:
@JavierArroyoBastida. I have tested a bit on the hydronic heat pump model, trying to implement a controller that modifies the temperature setpoint for the zone. However I am struggling a bit with the implementation due to how the setpoint overwrite block is implemented. 
As of now, the baseline controller adds an offset to the setpoint after the overwrite block (0.2 or 5.5K). At the same time, the setpoint overwrite blocks are limited to the comfort range. E.g. one can not lower the setpoint lower than 5.5 K above the comfort limit (288.15K) when there are no occupants. 

One solution would be to lower the minimum temperature allowed in the overwrite block.
Maybe a solution that is more understandable for the users is to insert an overwrite block between switch1 and conPI

JavierArroyoBastida:
That's a good point you're raising. Maybe we should consider both solutions actually. I think both are realistic and using either heating/cooling setpoints or directly the temperature setpoints depends on how a thermostat is configured. I'll come up with an implementation for this.

@dhblum FYI. Notice that building physiscs and baseline control will not change so it should be an easy edit.

haraldwalnum:
Thanks. Another more general note on the documentation for all the emulators, is that it would be nice to have a system sketch (P&ID), showing the connections and inputs and outputs. This could just be a screenshot of the modelica model, but make sure that the names of the input/output blocks are visible. 
Now it is quite hard to understand the model without opening it in Dymola, and I quess this should not be a prerequisite for those who want to test it.

dhblum:
I'm letting @icupeiro know of this potential change.

And @haraldwalnum your point on schematics is well-taken and I think would be good particularly as systems get more complex.

dhblum:
New, separate, issue for https://github.com/ibpsa/project1-boptest/issues/299#issuecomment-788838326 opened at #301.

PR #300, opened 2021-03-03
Issue299 bestest hp setpoint

JavierArroyoBastida:
For https://github.com/ibpsa/project1-boptest/issues/299. 
@haraldwalnum @icupeiro FYI.


dhblum:
I think just in reading the documentation, the differences between ``oveTSetHea_u``, ``oveTSetCoo_u``, and ``oveTSet_u`` is pretty confusing.  Is ``oveTSet_u`` not sufficient alone?  I think the offset in heating setpoint is part of the baseline controller, such that  ``oveTSetHea_u`` is unnecessary.  Then a user basically can override the setpoint given to the PI controller (which they could use to implement their own offset if they like based on the defined comfort region), or the heat pump actuation itself.  Also I think ``oveTSetCoo_u`` is unnecessary in general because there is no cooling system.  Thoughts?

JavierArroyoBastida:
Right now one can either overwrite the the bounds of the comfort range or the zone temperature setpoint directly. If there is confusion I think it's better to keep only the zone temperature setpoint overwrite (`oveTSet_u`) and remove `oveTSetHea_u` and `oveTSetCoo_u`.

dhblum:
I would vote to have only ``oveTSet_u``.

icupeiro:
I also got confused when setting up the MPC, and as David pointed there is no cooling system.
So I would also vote to have only `oveTSet_u`

JavierArroyoBastida:
Done! waiting for tests to pass.

Issue #304, opened 2021-03-26
Fix /name


dhblum:
@kbenne Pointed out that the ``/name`` api returns a string, though the flask api specifies application/json.  To make this correct and consistent with other api calls that do return jsons, this issue will change ``testcase.get_name()`` so that it returns a dictionary as ``{'name': <str>}``.

This issue will also take the opportunity to update the name that is returned to be that which will be defined in ``config.py`` for each test case.  Currently, it is returning the name of the .fmu, which is ``'wrapped'`` for all test cases.

dhblum:
Just another note that this is important for #282 and boptest-service api.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/305.

PR #305, opened 2021-03-26
Issue304 name

dhblum:
This is for #304.  In particular, it:

- ``testcase.get_name()`` returns dictionary
- ``config.py`` now contains test case name, which is used in ``testcase.get_name()``

@kbenne @JavierArroyoBastida FYI.  Should be a simple change and basically ready to merge once tests pass.  Let me know any objections.


JavierArroyoBastida:
Thanks for implementing this @dhblum. Looks good in general, although I see a few files being added in `testcases/multizone_office_simple_air`. Is it the intention to add them in this PR?

dhblum:
Thanks @JavierArroyoBastida for the catch!  Did NOT mean to add those.  I removed them.  Glad you double checked.

Issue #307, opened 2021-04-05
Correct bestest_air resources


dhblum:
The Resources and Modelica model are inconsistent with when the occupancy period ends.  The Modelica model schedules end occupancy AT the start of hour 18.  Meanwhile, the Resources data indicates that occupancy ends BY the start of hour 18, and due to the 10 minute sampling interval, causes the setpoint and internal gain forecasts to being changing 10 minutes too early.  This issue is to make the Resources data indicate occupancy ends AT the start of hour 18.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/308.

PR #308, opened 2021-04-05
Issue307 bestest air resources

dhblum:
This is for #307.


dhblum:
@JavierArroyoBastida This is a specific fix for bestest_air (description in the issue thread), let me know if you have concerns merging.

JavierArroyoBastida:
Thanks for spotting and fixing the issue. I'm doubting now whether the hydronic cases would require a related update as well. After your fix, the `.csv` file of the resources folder of the bestest air looks as follows for a step "activation":

image

Where the first column is the time in seconds, the second the time in hours, and the third the number of occupants (same reasoning applies for other variables related to occupancy). And for a step "deactivation":

image

Meaning that the event time always holds the value of the active period, which I think makes sense. 

For the bestest hydronic simple case we get the following for a step activation:

image

And for a step deactivation:

image

And similarly happens for the bestest_hydronic case with heat pump. 
Meaning that the hydronic cases follow a backward fill interpolation for the event times, instead of holding the values during the activation periods. I suggest to close this PR (I think your fix looks good) and adopt the convention from now on that event times should hold values during active step periods, for which I'll open a new issue to update the hydronic cases.

dhblum:
@JavierArroyoBastida That sounds good to me.  Thanks for spotting on the hydronic cases.

PR #314, opened 2021-04-19
Issue270 scenario time 2

dhblum:
This is for #270.  See change tracking here: https://github.com/ibpsa/project1-boptest/issues/270#issuecomment-781697871.  In general, this PR aims to be the final one for initial implementation of the ``/scenario`` API option ``'time_period'``.  Final to-dos:

- [x] Fix merge conflicts
- [x] Review
- [x] Update wrapped fmus to contain days.json
- [x] Fix bestest_hydronic can't have peak_heat_day as day 7, since need 1 week warmup + 1 week before day 7.


testing/references/bestest_hydronic/plot_references.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python2
```
JavierArroyoBastida, line 1: I think it'd be better to add this kind of functionality in the `testing/utilities.py` script and to remove this one from this PR. I have a script that does exactly this and is general for all test cases. I'll raise an issue for this.

testing/references/bestest_hydronic_heat_pump/plot_references.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python2
```
JavierArroyoBastida, line 1: Same comment as with the `bestest_hydronic` case

JavierArroyoBastida (approved):
Great additions @dhblum, thanks a lot for this. I'd just suggest to remove the `plot_references.py` scripts from this PR.

testing/references/bestest_hydronic/plot_references.py
dhblum: Thanks for the catch.  Yes, didn't mean to include.

testing/references/bestest_hydronic_heat_pump/plot_references.py
dhblum: Thanks for the catch.  Yes, didn't mean to include.

dhblum: Done in https://github.com/ibpsa/project1-boptest/pull/314/commits/d8851fb75af5911cfdf2d089e8235238b2f71ac0.

testing/references/bestest_hydronic/plot_references.py
dhblum: done in https://github.com/ibpsa/project1-boptest/pull/314/commits/d8851fb75af5911cfdf2d089e8235238b2f71ac0.

dhblum:
@JavierArroyoBastida Added to to-do a fix needed for bestest_hydronic peak heat day.  Working on it.

dhblum:
@JavierArroyoBastida please check https://github.com/ibpsa/project1-boptest/pull/314/commits/0d14502dfb8edab9ace7743d837431a74053abb2.  In it, the find_days algorithm is updated so that it uses the whole year to calculate the median, but when choosing the day only uses the feasible days.

JavierArroyoBastida:
I think https://github.com/ibpsa/project1-boptest/commit/0d14502dfb8edab9ace7743d837431a74053abb2 makes sense, just added one more [comment ](https://github.com/ibpsa/project1-boptest/pull/314#commitcomment-49758548) to it.

dhblum:
Updating reference results and will recompile wrapped fmus.

Issue #315, opened 2021-04-20
Compare references


JavierArroyoBastida:
For every PR involving changes in references (almost all PRs) I always end up following the same process:
1. Implement the changes.
2. Change the name of the folder with references that may be affected by the changes. 
3. Run the tests and let the new references be created (this happens automatically if references folder can't be found, which is the case since name has been changed in 2.). 
4. Compare old set of references to the new ones, mainly by plotting the most relevant variables. 

For the latter step I've eventually come up with a script that visually shows the changes between the old and the new references. I think it may be useful for other developers, so this is to propose the addition of this functionality to the `testing/utilities.py` script.

dhblum:
I do similar.  One thing I've thought about is doing something similar to what is done in the Modelica Buildings Library.  It uses [BuildingsPy](https://github.com/lbl-srg/BuildingsPy) for regression tests and the legacy version used to generate plots showing the reference and test data automatically if the test failed.  Currently, it uses [pyfunnel](https://github.com/lbl-srg/funnel) for the tests and think it still may do something similar.

JavierArroyoBastida:
I'm not that familiar with BuildingsPy but it sounds like the right way to do it. What do you think of going with https://github.com/ibpsa/project1-boptest/pull/316 for the moment and implement the upgrade to BulidingsPy and pyfunnel when we get the chance?

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/316.

PR #316, opened 2021-04-20
Issue315 compare references

JavierArroyoBastida:
This is for https://github.com/ibpsa/project1-boptest/issues/315.


dhblum:
I have a bit of concern with putting the ``compare_references.py`` into the testcase directories, which may be confused with BOPTEST testing and not unit testing.  What do you think about collapsing ``compare_references.py`` into one file with many classes or methods with a ``__main__`` where we can just specify the test case name, and putting the file within ``/testing``.

JavierArroyoBastida:
That's a very good point. I've moved `compare_references.py` in this [commit](https://github.com/ibpsa/project1-boptest/pull/316/commits/79de3beab971b082f286e495d2738e4619b67fa1). Let me know if that's what you had in mind.

dhblum:
Yeup, that's just what I had in mind.  Thanks!  I'll merge.

dhblum (approved):

Issue #313, opened 2021-04-14
Add scenario period documentation


JavierArroyoBastida:
This issue is to add the scenario period documentation to the bestest hydronic cases following https://github.com/ibpsa/project1-boptest/issues/270#issuecomment-819130077.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/323.

Issue #312, opened 2021-04-14
Update Uccle.TMY


JavierArroyoBastida:
Both bestest hydronic cases use the Uccle.TMY weather file. This file is expected to be modified soon because the solar irradiation has been found to be shifted by one hour, see: https://github.com/open-ideas/IDEAS/issues/1206. This issue is to update the cases accordingly.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/320.

PR #320, opened 2021-04-23
Issue312 update Uccle TMY

JavierArroyoBastida:
This is for https://github.com/ibpsa/project1-boptest/issues/312. This PR changes the weather file of the bestest hydronic cases so that the new cases are located in Brussels instead of Uccle. This change affects the testing periods specified in `days.json`, and also the reference results are radically affected. I've tried to make sure that the underlying behavior remains the same though. 

Here you can see 20 simulation days for the bestest hydronic simple case when using the new weather file in Brussels (simulation 1) and the old weather file in Uccle (simulation 2):
image

And for the bestes hydronic case with a heat pump:
image

I'll update the cases documentation using [this issue](https://github.com/ibpsa/project1-boptest/issues/313) once this is merged.


dhblum (approved):
This looks good to me.  Thanks for taking care of it.  I read the IDEAS issue tracker for the reason for using the new file instead of fixing the old.  Makes sense.

I understand reasoning for changing the documentation all at once in https://github.com/ibpsa/project1-boptest/issues/313 since you are working across two repositories, but please do this as soon as you can after merging this so we don't have a discrepancy for too long of where the documentation says this test case is located in terms of climate.

PR #323, opened 2021-04-23
Issue313 scenario period doc

JavierArroyoBastida:
For https://github.com/ibpsa/project1-boptest/issues/313.


dhblum (changes requested):
This looks good, thanks.  Do you mind just adding the ``/scenario`` API specifier for the electricity prices as well as in https://github.com/ibpsa/project1-boptest/blob/master/testcases/bestest_air/doc/index.html#L475?

JavierArroyoBastida:
@dhblum I'm not sure why tests are not passing. The build shows:
image

According to the indicated [website ](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) the problem may be related to one of the `multizone_residential_hydronic` taking longer than 10 minutes. It can be solved using `travis_wait` before the command to run the tests, but I wonder why this is only happening now and not in the PR of this testcase. Any idea why?

dhblum:
It's a good question.  I don't know the answer, except that maybe if the length of the test is very close to being above or below the 10 minute mark, if it takes a few extra seconds it trips.  I restarted the build for now to see if it happens twice in a row.  May be an issue too with more complex emulators.

dhblum (approved):
Looks good!  Thank you!

Issue #302, opened 2021-03-19
baseline step size


JavierArroyoBastida:
Current master returns slightly different baseline KPI results when using different step sizes for the same scenario and simulation periods. Here the script to reproduce the issue:

[test_kpis_baseline.txt](https://github.com/ibpsa/project1-boptest/files/6170811/test_kpis_baseline.txt)

Running this script in master and in https://github.com/ibpsa/project1-boptest/tree/issue270_scenarioTime_2 returns in both cases the exact same results:

```
tdis_tot
8.29017550098
8.35305147656
cost_tot
167.731006157
167.730361748
```

Where it's shown a small discrepancy when running the baseline using a step size of 12 hours vs. a step size of 24 hours. 
In the beginning I thought it could be a bug from the KPI calculation module but after reviewing it I figured out that this is because the communication points returned from pyfmi depend on the simulation length, which in turn affects the integration to calculate KPIs. This issue can be solved directly by merging https://github.com/ibpsa/project1-boptest/tree/issue240_memory to master, which fixes the number of communication points. After so the script returns:

```
tdis_tot
8.28941468357
8.28941468357
cost_tot
167.741309909
167.741309909
```
which proves that baseline KPIs are independent of step size. 
My suggestion is to prioritize merging not only https://github.com/ibpsa/project1-boptest/tree/issue270_scenarioTime_2 but also https://github.com/ibpsa/project1-boptest/tree/issue240_memory. 

EDIT: I forgot to mention that I used the `bestest_hydronic_heat_pump` case for all tests.

JavierArroyoBastida:
Below the branch that I used to merge https://github.com/ibpsa/project1-boptest/tree/issue240_memory and that leads to same baseline results independently of the step size:

https://github.com/JavierArroyoBastida/project1-boptest/tree/issue302_baselineStepSize

dhblum:
Thanks for reporting this.  Indeed I reproduced your results, error and fix, on my machine.  I merged latest master to https://github.com/ibpsa/project1-boptest/tree/issue240_memory and pushed just to have tests run.  

But https://github.com/ibpsa/project1-boptest/tree/issue240_memory makes it so that all trajectories are stored, and reported, at 30s intervals, for instance when using /results.  This could be more akin to how sensors typically out at regular intervals.  One issue is if a user wants to test at a smaller step size, maybe for a feedback controller.

JavierArroyoBastida:
[...]

PR #324, opened 2021-04-26
Issue302 baseline step size

JavierArroyoBastida:
This is for https://github.com/ibpsa/project1-boptest/issues/302 and would also close https://github.com/ibpsa/project1-boptest/issues/240.


JavierArroyoBastida:
Test case references require an update, which was expected... I'll wait for https://github.com/ibpsa/project1-boptest/pull/192.

testcase.py
```diff
@@ -97,6 +99,8 @@ def __initilize_data(self):
         for key in self.input_names:
             self.u[key] = np.array([])
         self.u_store = copy.deepcopy(self.u)
+        # Results filtering for pyfmi
+        self.options['filter'] = self.output_names + self.input_names
```
dhblum, line 103: Should the filter option be set in the __init__ with the other options?  This just may be more consistent to have all the options set in the __init__ when they can be (basically move this to after line 60, and move call of __initialize_data accordingly in __init__).

dhblum (changes requested):
I think this looks good in general, just the one comment.

dhblum:
Just a note too, the latest travis check on this PR took about 30 minutes less time than the travis check from the last merged PR, and the reference results directory in this PR is half the size of the one currently on maser.  Good stuff!

testcase.py
JavierArroyoBastida: I think that totally makes sense. Implemented in https://github.com/ibpsa/project1-boptest/pull/324/commits/b9110650339c09bf2e80c00cbb7d04b711760a41.

JavierArroyoBastida:
That's great news indeed!

dhblum (approved):
Looks good!  Will merge.  Thanks again.

Issue #325, opened 2021-05-04
Bestest hydronic fmus


JavierArroyoBastida:
Looks like in https://github.com/ibpsa/project1-boptest/pull/323 I did not update the wrapped fmus of the hydronic cases with the right data that was fixed in https://github.com/ibpsa/project1-boptest/pull/322. This is to update these fmus.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/326.

Issue #183, opened 2020-04-14
Compilation of test cases in unit-tests


JavierArroyoBastida:
@dhblum it rings a bell that we've discussed this already in the past but I think we might need to give a deeper though. It's about the compilation of the test cases that takes place in the beginning of most of the unit-tests. I think in some of them it makes sense like for `test_parser` or `test_data` as their functionality is involved in the compilation process. For the others I'm not sure if it's needed, or even if it's a good practice. Specifically for `test_kpis` I doubt that it's needed to compile the models for unit-testing as the tests are reading data from a deployed test case. 

An example of why it might not be a good practice arises in commit https://github.com/ibpsa/project1-boptest/commit/a9b9754595dde840914db2aa39f1ffbefa8d5cb2 where we forgot to update `wrapped.fmu` and `wrapped.mo` that should had included the latest KPI tags with CO2 measurements. Still all tests were being passed because the original `.mo` models did have the tags and these were compiled in every unit-test, resulting in the right `wrapped.fmu` that was used for the unit test but did not persist afterwards. This has been solved in https://github.com/ibpsa/project1-boptest/commit/5727f6b4a20c0b0d589502a37798397e95991b6c. 

I understand that unit-tests need to be independent from each other. Precisely for that I think we should avoid compilation of test cases in each of the unit-tests and rather accept that some unit-tests are built around a compiled `wrapped.fmu` that should include the latest features. What do you think?

dhblum:
Thanks for raising the point.  It is a good one.  I agree that testing the test case fmus as made available in the repo should be added, and that ``test_kpis`` may not need to recompile test cases if it only reads in data.  

But one thing I'm thinking should still be tested somehow is that the Modelica models for the test cases that are included in the repo should be tested to match what is made available in the associated fmus.

JavierArroyoBastida:
I agree that that is a necessary test. Actually such a test would have triggered an error with the issue that I was raising in my previous comment: https://github.com/ibpsa/project1-boptest/issues/183#issue-599447394. However, I think we don't get that check by just compiling the models and using the newly compiled model in the tests because that process does not use at any point the fmu as made available in the repo. A test I can think of is, for each provided test case:
1. Compile the Modelica model of the repo into fmu_test (this is what we are actually doing in the beginning of each test). 
2. Load fmu_test
3. Load the fmu available in the repo as fmu_repo
4. Assert that (some of) the simulation result trajectories of fmu_test and fmu_repo coincide

[...]

PR #326, opened 2021-05-04
Issue325 bestest hydronic fmus

JavierArroyoBastida:
For https://github.com/ibpsa/project1-boptest/issues/325. @dhblum please let me know what you'd like to merge first if this one or https://github.com/ibpsa/project1-boptest/pull/324, to avoid updating references twice. I think it makes sense to do https://github.com/ibpsa/project1-boptest/pull/324 first since references are updated there already, in that case I'd get this one ready inmediately after.


dhblum:
What you suggested makes sense.  I just merged #324. Also, I suppose this is another instance of needing to address https://github.com/ibpsa/project1-boptest/issues/183.

JavierArroyoBastida:
Thanks for meging https://github.com/ibpsa/project1-boptest/pull/324. I'm going to update references here now. Also good point that you're raising about https://github.com/ibpsa/project1-boptest/issues/183, it has never been a priority but could certainly help in some cases like this one.

JavierArroyoBastida:
@dhblum it has taken me a while to understand why forecast-related references have not needed update here, but in fact it has happened what was explained in https://github.com/ibpsa/project1-boptest/issues/183: references have been right all along even though the `wrapped.fmu` was using the wrong data. Tests would still pass because the right `wrapped.fmu` is compiled and used in the tests. Notice that in https://github.com/ibpsa/project1-boptest/pull/326/commits/7e0e19267270b4682df684368f0abfb6eca9b337 only `time_rat` needed update. So if that all makes sense and test pass this should be ready to merge too.

dhblum:
Yeup, that makes sense, and is why I brought up #183.  Thanks for patching, I'll merge.

dhblum (approved):

Issue #327, opened 2021-05-12
Computational Time Ratio KPI Calculation


dhblum:
In the Design Requirements and Guide, Computational Time Ratio is defined to allow for variable control step size.  The API also allows a user to change a step size mid simulation.  However, the implementation of the KPI calculator ([here](https://github.com/ibpsa/project1-boptest/blob/538acebc318e1d5461434cdef1e883ef3e3b4a8c/kpis/kpi_calculator.py#L520)) uses the ``testcase.step`` attribute, which would just be the last step size used, but not a record of those used through the duration of the simulation.  I think this should be rectified by storing the step sizes so the Design Requirements and Guide spec and API implementation can be satisfied.

dhblum:
Currently, the Design Requirements and Guide has computational time ratio defined as the average computational time divided by the average simulation step:

<img width="335" alt="Screen Shot 2021-05-13 at 5 18 49 PM" src="https://user-images.githubusercontent.com/19480999/118189113-5e07a300-b40f-11eb-88f6-b86cc276ecb6.png">

I think is is incorrect to say that sum(a) / sum(b) = sum(a/b).  For instance, consider 10 steps where if at every step the controller takes 1s and the simulation step is 10 s, then the value of the LHS = 0.1 where the value of the RHS = 1.0.

I think instead this KPI is meant to take the average ratio of t_c and T_s over all time steps k, which would require dividing the RHS by n.

dhblum:
Other thoughts welcome, FYI @SenHuang19 @JavierArroyoBastida.

SenHuang19:
I totall agree with you about " sum(a) / sum(b) != sum(a/b)"
I suggest we either do the average or include both mean and std

SenHuang19:
Besides, should we consider if a controller fails to calculate the output in time, we have to sustain the value of control signals from a previous successful run?

dhblum:
I agree with sticking with the average right now, and not also including std.

Regarding if the controller fails, since the current implementation is controller-blocked (e.g. the controller needs to advance time when ready), it is up to the controller to identify it has failed and advance time with the previous successful run.

JavierArroyoBastida:
I totally agree that we should rectify this to take the average ratio of t_c and T_s over all time steps k. I'd also avoid std just to not complicate things since so far every KPI has a single scalar result associated to it. I think the implementation should be quite simple, I'll let you know if I get started with this. Please do the same if you do so that we don't overlap.

[...]

PR #329, opened 2021-05-14
Issue327 com tim rat

dhblum:
This is for #327.

@JavierArroyoBastida After you review, I'll update reference results.  Please also let me know if it is worth keeping the plot functionality of ``kpi_calculator.get_computational_time_ratio()``, as it would need to be fixed to use a vector of stored simulation steps, rather than assuming constant throughout the test period.

### DIFF
--- a/testcase.py
+++ b/testcase.py
@@ -68,7 +68,7 @@ class TestCase(object):
         self.start_time = 0
         self.initialize_fmu = True
         self.options['initialize'] = self.initialize_fmu
-        self.elapsed_control_time = []
+        self.elapsed_control_time_ratio = []
         # Instantiate a KPI calculator for the test case
         self.cal = KPI_Calculator(testcase=self)
         # Set default scenario
@@ -196,7 +196,7 @@ class TestCase(object):
         # Calculate and store the elapsed time
         if hasattr(self, 'tic_time'):
             self.tac_time = time.time()
-            self.elapsed_control_time.append(self.tac_time-self.tic_time)
+            self.elapsed_control_time_ratio.append((self.tac_time-self.tic_time)/self.step)
 
         # Set final time
         self.final_time = self.start_time + self.step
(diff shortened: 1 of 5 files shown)


testcase.py
```diff
@@ -566,7 +566,7 @@ def get_name(self):
 
         return name
 
-    def get_elapsed_control_time(self):
+    def get_elapsed_control_time_ratio(self):
         '''Returns the elapsed control time vector for the case.
```
JavierArroyoBastida, line 570:
```suggestion
        '''Returns the elapsed control time ratio vector for the case.
```

testing/test_testcase1.py
```diff
@@ -277,6 +278,33 @@ def test_return(self):
...
+
+    def test_kpi(self):
+        '''Tests the calculation of the kpi.
+
+        '''
+
+        # Run test
+        requests.put('{0}/initialize'.format(self.url), data={'start_time':0, 'warmup_period':0})
+        step = requests.get('{0}/step'.format(self.url)).json()
+        for i in range(5):
```
JavierArroyoBastida, line 301: Should we test that we can also handle variable control steps for computational time ratio? I mean, we could set a variable step as function of the `i` index either in this same test or [here](https://github.com/ibpsa/project1-boptest/blob/bf9d49873ec3b475c9324092141628b32ea9c785/testing/test_kpis.py#L131).

JavierArroyoBastida (commented):
@dhblum this is looking good, thanks for the implementation. Just added two suggestions.

testing/test_testcase1.py
dhblum: Yes, good point, thanks.  I can implement.

dhblum: Added in https://github.com/ibpsa/project1-boptest/pull/329/commits/d470b43388002cb96568d2f416154736f60e8a25.

JavierArroyoBastida: Thanks @dhblum! looking good.

JavierArroyoBastida (approved):

Issue #237, opened 2020-09-30
Normalize energy usage by floor area


dhblum:
This issue is to address the normalization of energy usage by floor area of the test case, which could make it possible to compare results with other case studies on similar buildings, or potentially across building types.  Though, different building types will naturally have different EUIs within a likely range.

JavierArroyoBastida:
I like this initiative. Is the idea to compute the EUI as an additional core KPI? or substituting of the currently used `ener_tot`? or just as another KPI not belonging to the core set of KPIs?

mwetter:
I would rather substitute the energy consumption (kWh) with normalized energy consumption (kWh/m2). You could separately track the floor area so you can see which buildings are within a certain size, but comparing kWh across buildings with different floor area is not meaningful. This is why I think we should be using kWh/m2.

dhblum:
@mwetter I wanted to open this discussion again to point out that our test scenarios now are not annual tests, but two-week periods at the moment.  Perhaps this gets expanded in the future.  But my notion of kWh/m2 is that it is usually associated with annual energy metrics and I have a bit of concern that reporting normalized by floor area may add some confusion when the numbers are not comparable to EUIs for similar building types found elsewhere, without further understanding that they are not meant for annual EUI comparison.  Perhaps this is of low concern, but wanted to see what others think.

I also want to raise the question of if cost, discomfort, and CO2 emission core kpis should also be normalized by floor area?

mwetter:
@dhblum : I still think it makes sense to normalize by the floor area to remove the dependency on the arbitrary size of the building. One way to avoid confusion with EUI would be to report kWh/(m2.a) for annual, as it is done in some energy codes, and maybe kWh/m2 otherwise. Or if one has multiple 2 weeks periods in different seasons, maybe a weighted sum could be used to extrapolate from short-term measurements to expected annual performance, but this would need careful design and/or research.

JavierArroyoBastida:
@mwetter is your idea to normalize by the floor area only the energy use? or are you thinking of normalizing also others like operational cost, emissions, thermal discomfort, IAQ, etc. ?

mwetter:
@JavierArroyoBastida I suggest to normalize everything that scales with floor area, such as cost and emission. However, thermal comfort should be an average (or max discomfort) across all zones, maybe weighted by floor area and/or importance so that a large office zone may get more weight than a small hallway.

[...]

PR #330, opened 2021-05-28
Issue237 kpi normalization

dhblum:
This is for #237.  In particular, it:

- Adds ``area`` as argument to ``config.py``
- Uses floor area to normalize energy, cost, and emission kpis
- Uses number of conditioned zones to normalize thermal and iaq discomfort
- Updates unit test reference results


dhblum:
@JavierArroyoBastida I think this is ready to review if there are no other comments in #237.

testcases/multizone_residential_hydronic/config.py
```diff
@@ -25,6 +26,7 @@ def get_config():
     # Enter configuration information
     'name'     : 'multizone_residential_hydronic',
     'fmupath'  : 'models/wrapped.fmu',
+    'area'     : 8.3*14.4,
```
JavierArroyoBastida, line 29: Shouldn't we remove the area of the garage here to use only the area of conditioned zones? It'd be: `8.3*14.4-20.52` I think it's more fair since thermal discomfort and indoor air quality are also computed based on those zones only.

JavierArroyoBastida: Also I'm double checking the lengths of the building and always get a length of 12.7m and a width of 8m. This is based on the specification of the lengths for each zone and orientation at the beginning of the model. So if I'm not mistaken I think it should actually be `8*12.7-20.52`. I'm not sure where the 8.3m and 14.4m are coming from, I'm not able to reproduce those values again. I think it'd be better if we update the figure of the documentation accordingly, I can open a new issue for that.

JavierArroyoBastida (changes requested):
Thanks a lot for getting this done @dhblum. Very clean implementation, I've got just a couple of minor comments.

kpis/kpi_calculator.py
```diff
@@ -223,8 +223,8 @@ def get_thermal_discomfort(self, plot=False):
                 self.tdis_dict[signal[:-1]+'dTupper_y'] += \
                     trapz(dT_upper,self._get_data_from_last_index('time',self.i_last_tdis))/3600.
                 self.tdis_tot = self.tdis_tot + \
-                    self.tdis_dict[signal[:-1]+'dTlower_y'] + \
-                    self.tdis_dict[signal[:-1]+'dTupper_y']
+                    self.tdis_dict[signal[:-1]+'dTlower_y']/len(self.sources_idis) + \
```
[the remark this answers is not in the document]
dhblum: Yes good catch.  I think this is just a typo by me.

testcases/bestest_hydronic/config.py
```diff
@@ -26,6 +27,7 @@ def get_config():
     # Enter configuration information
     'name'     : 'bestest_hydronic',
     'fmupath'  : 'models/wrapped.fmu',
+    'area'     : 48.0,
```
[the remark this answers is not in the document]
dhblum: Sure that's fine too.

testcases/multizone_residential_hydronic/config.py
dhblum: > Shouldn't we remove the area of the garage here to use only the area of conditioned zones? It'd be: 8.3*14.4-20.52 I think it's more fair since thermal discomfort and indoor air quality are also computed based on those zones only.

You bring up a couple of clarifying points.  I would suggest that the floor area used be total floor area, not only conditioned floor area, since this is what is used in [EnergyStar](https://www.energystar.gov/buildings/benchmark/understand_metrics/what_eui) and [CBECS](https://www.eia.gov/consumption/commercial/data/2012/c&e/cfm/c1.php) EUI reporting.  The question remains to me though if a garage should be considered building floor area?  I'd say probably not, since a garage could be detached, or it may not be included within the insulated envelope of the building.

> Also I'm double checking the lengths of the building and always get a length of 12.7m and a width of 8m. This is based on the specification of the lengths for each zone and orientation at the beginning of the model. So if I'm not mistaken I think it should actually be 8*12.7-20.52. I'm not sure where the 8.3m and 14.4m are coming from, I'm not able to reproduce those values again. I think it'd be better if we update the figure of the documentation accordingly, I can open a new issue for that.

I took the numbers 8.3*14.4 from the schematic in the documentation and did not check the actual modelica model.  They also match the text in the documentation that states the floor area is approximately 120 m^2.  Both should match what is actually in the Modelica model, and specifying what the floor area is of the garage and whether or not it is included in the area calculation for EUI, cost, emissions would be good.

kpis/kpi_calculator.py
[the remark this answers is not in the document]
dhblum: Fixed in https://github.com/ibpsa/project1-boptest/pull/330/commits/fb3c8c6a2e73a1b8d6f24a0772e9ba7848696747.

testcases/bestest_hydronic/config.py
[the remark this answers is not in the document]
dhblum: Fixed in https://github.com/ibpsa/project1-boptest/pull/330/commits/c274fc573e1521106aab2d9aa7857d8f3bcf17f0.

testcases/multizone_residential_hydronic/config.py
JavierArroyoBastida: I suggest to merge https://github.com/ibpsa/project1-boptest/compare/master...JavierArroyoBastida:issue333_correctSchema?expand=1 first, and use `12.7*8-20.52` as we don't consider garage as part of the floor area. Notice though that if the attic is to be considered floor area and we decide to inclue unconditioned zones then the attic floor should count as well and the area to be used would be: `12.7*8*2-20.52`. I still think it makes more sense to account only for conditioned spaces, but I guess it's a bit subjective like our discussion in https://github.com/ibpsa/project1-boptest/issues/237 and I don't have a strong opinion on it so @dhblum feel free to decide on what's best.

dhblum: Per my suggestion [here](https://github.com/ibpsa/project1-boptest/issues/237#issuecomment-856822390), the unconditioned attic and unconditioned garage should NOT be included in the floor area.  Please make sure changes in #334 are consistent with this.  I would specify in the documentation something like, "the area not including the unconditioned attic and unconditioned garage is <area>."

JavierArroyoBastida: Sounds good. I've stated that in https://github.com/ibpsa/project1-boptest/pull/334/commits/62336503a523e440042500a803efb5156cf476dd.

dhblum: Cool thanks.  I've updated the area in ``config.py`` and unit test reference results in https://github.com/ibpsa/project1-boptest/pull/330/commits/54ea93b93546d41f448c34e95ada7e187d7b5422.

dhblum:
Note that I've updated and cleaned/clarified kpi documentation in https://github.com/ibpsa/project1-boptest/pull/330/commits/9aa3761137751939f6a4b797b8d6dd7b0499ab0a.

dhblum:
@JavierArroyoBastida Feel free to complete the review and let me know if you detect additional issues.

testcases/multizone_residential_hydronic/config.py
dhblum: #334 has been merged and this branch updated with latest master.

JavierArroyoBastida (changes requested):
@dhblum, the new changes are looking good to me, but one of the builds has failed for some reason. I don't seem to identify why. Do you have an idea?

dhblum:
I checked out the log and it's a pretty cryptic error.  Odd to that one would pass and the other fail.  Maybe something with Travis servers?  I restarted the build so let's see if it passes this time just to be sure.

JavierArroyoBastida:
Good, I was going to propose the same unless you could think of a clear reason.

dhblum:
Oy, now we're getting pull rate limits on docker containers...

dhblum:
I don't know the difference between the builds.  But looks like both are passing now!

JavierArroyoBastida:
Nice to see it works!  It looks like we are not the only ones having only one of the builds breaking because of a random infrastructure-related issue: see https://github.com/travis-ci/travis-ci/issues/3241#issuecomment-230056915 and the whole issue around. As I understand, the branch build tests the head commit of a branch, and a PR build tests the commit as if merged to the master. I think the first one may be redundant, and we could only use the PR build, which I think is what really matters, for a more efficient use of resources. @mwetter, what do you think? [This link](https://stackoverflow.com/questions/31882306/how-to-configure-travis-ci-to-build-pull-requests-merges-to-master-w-o-redunda) explains how this can be done.

JavierArroyoBastida (approved):

dhblum:
Thanks @JavierArroyoBastida for the info on the travis builds.  I like having the head commit tested too, though I think this also is done with each commit that gets pushed to the branch, so indeed may be redundant.

Issue #335, opened 2021-06-11
Prep for Version Release


dhblum:
As BOPTEST becomes used it will be important to version to allow for clear benchmarking.  This issue is to prep for an initial version release with the software that is largely available now on the master branch.

I suggest the first version to be 0.1.0, indicating some features are still in development.  Updates that include "major" changes (such as API changes, fixes that make major changes to results, or new emulator availability) would bump the second decimal and updates that include "minor" changes (such as small bug fixes, performance updates, documentation updates) would bump the third.

dhblum:
Initializing a suggested to-do, which may be updated:

- [ ] Add attribute ``testcase.version`` and API endpoint ``version`` as GET request.
- [ ] Update schematics per https://github.com/ibpsa/project1-boptest/issues/301.
- [ ] Remove ``/template`` and ``api_example.py`` since they are outdated.
- [ ] Test that ``examples/python`` run in Python 3 as in https://github.com/ibpsa/project1-boptest/issues/328

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/336.

Issue #328, opened 2021-05-12
Make examples easier to use and fix bug


FWuellhorst:
Besides the following minor bug (python 3.8), the example for testcase1 works fine:
Line: 
`points = measurements.keys() + inputs.keys()`
should be 
`points = list(measurements.keys()) + list(inputs.keys())`, at least in python 3.8. In python 2.X, both lines work.

Additionally, users could skip one step if the file would start with 
```python
import sys
import pathlib
sys.path.insert(0, str(pathlib.Path(__file__).absolute().parents[2]))
import requests
import numpy as np
from examples.python.custom_kpi import custom_kpi_calculator as kpicalculation
```
instead of 
```python
import requests
import numpy as np
from examples.python.custom_kpi import custom_kpi_calculator as kpicalculation
```

I would be happy to open a pull request with listed changes and check if other python examples are affected as well.
If the solutions don't align with the project, feel free to close the issue.

JavierArroyoBastida:
@FWuellhorst I like the edit to have the example working with both python 3.8 and 2.x. A PR will be very welcome for that. Note that you can check if changes affet to other examples and modules in general by running the unit tests locally using the `make test_all` command in the testing directory. This takes quite long (one hour or so) but ensures that all other functionality is not affected. If you are sure that your changes are only affecting one module then you can check only that specific part by calling one of the following:
```
make test_parser
make test_data
make test_forecast
make test_kpis
make test_testcase1
make test_testcase2
make test_testcase3
make test_bestest_air
make test_bestest_hydronic
make test_bestest_hydronic_heat_pump
make test_multizone_residential_hydronic
```
The Travis tests will always run all tests anyways.

I'm not sure the other edit is required though since the examples assume that those directories can already be found by the environment.

FWuellhorst:
Ok, perfect! I will try the tests out. Thanks for the information. I will create a PR soon.

Regarding the PYTHONPATH: Yes, I know the tutorial and ReadMe include this step. The change could be a possible enhancement to reduce the number of steps necessary to test BOPTEST. However, as users have to install docker, setting the PYTHONPATH should not be a problem.

dhblum:
Agreed, thank you @FWuellhorst!  One question is if ``sys.path.insert(0, str(pathlib.Path(__file__).absolute().parents[2]))`` edits the PYTHONPATH permanently or just for as long as the process runs?

[...]

PR #229, opened 2020-08-06
Python3 support, except parser

dhblum:
For #146 and #228 

- Python3 support and pyfmi from anaconda in BOPTEST docker container.  Removes dependence on JModelica.org BOPTEST docker container.
- Python3 support for all repo modules, except ``parser/parser.py`` and ``data/data_generator.py``.  These still require Python2 due to JModelica.org dependence for compiling FMUs.  Therefore, their unittests (``testing/makefile test_parser`` and ``testing/makefile test_data_generator``) also still must be run in ``jm`` testing docker image, which contains JModelica.org and Python2.
- Changes config.py to config.json.

PR #336, opened 2021-06-23
Issue335 ver rel

dhblum:
This is for #335 and #328 and #228 and #340 and #341.


dhblum:
With https://github.com/ibpsa/project1-boptest/pull/336/commits/333fa07534319c933a6c1fceb718cd74020b3254 and https://github.com/ibpsa/project1-boptest/pull/336/commits/2be8165a9d790eb1aa1cb54e08b474c00a1dc126, this PR also addressed #328.

testcase.py
```diff
@@ -25,6 +25,8 @@ def __init__(self):
 
         '''
 
+        # Set BOPTEST version number
+        self.version = '0.0.0'
```
SenHuang19, line 29: should we put the "version" into a __init__.py?

dhblum: What's the advantage(s) of that?

testing/utilities.py
```diff
@@ -416,6 +418,18 @@ class partialTestAPI(partialChecks):
 
     '''
 
+    def test_get_version(self):
+        '''Test getting the version of BOPTEST.
+
+        '''
+
+        version = requests.get('{0}/version'.format(self.url)).json()
+        r = re.compile('\d.\d.\d')
```
SenHuang19, line 427: maybe add some comments here for maintenance purposes

.travis.yml
```diff
@@ -19,5 +16,12 @@ install:
...
-script:
-  - (cd testing && make test_all)
+matrix:
+  include:
+    - python: 2.7
+    script:
+      - (cd testing && make test_all)
+
+    - python: 3.9
+    script:
```
SenHuang19: should we make the python3 as the primary script?

testing/utilities.py
dhblum: Done in https://github.com/ibpsa/project1-boptest/pull/336/commits/9955ba4fcfbf3596bb050a56d3ca90e9508a4556.

.travis.yml
dhblum: The modules that work within the boptest docker are Python 2.7 and are tested within a 2.7 environment (/data, /kpis, /parser).  The code that would utilize Python 3 if it is used for the main testing script is the example controllers and the unit testing code for the example controllers and the test cases.  The unit testing code could be updated to also work with Python 3, as in https://github.com/ibpsa/project1-boptest/pull/229/commits/29d664cd5d088b823bdc2dc6a0f70fc0f00db11c.  I think the only consequence would be the need to update reference results since there could be some numeric differences with Python 3 versions of pandas, numpy, etc.

I've made the updates to the unit testing code and am testing now to see if this is true.

dhblum: Indeed, reference results for timeseries just need to be updated for testcase1 and testcase3 for small numeric differences.  And two results in bestest_hydronic_heat_pump need to be updated due to python3 seemingly to order a list differently in two cases.

testcase.py
SenHuang19: I thought this was somehow a standard? https://packaging.python.org/guides/single-sourcing-package-version/#single-sourcing-the-version

dhblum: I see thanks.  What about 4. on the list in the link you provide?  Since the BOPTEST repo does not just consist of Python modules and is not "installed" via pip.  We could create a version.txt in the root directory and have testcase.py read it on instantiation?

JavierArroyoBastida (commented):
So far everything is looking good to me.

.travis.yml
```diff
@@ -14,10 +11,14 @@ services:
...
 
 before_script:
   - export PYTHONPATH=$PYTHONPATH:$(pwd)
 
-script:
-  - (cd testing && make test_all)
+jobs:
+  include:
+  - python: 2.7
+    script: cd testing && make test_all
```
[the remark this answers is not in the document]
dhblum: I'm not sure I understand what you mean.  As is written, all tests are run via ``test_all``.  The ``test_python3`` is only for testing that the examples also run in python3.  We could run ``test_all`` in python3, see my comment to @SenHuang19 above.

testcase.py
SenHuang19: looks good to me

.travis.yml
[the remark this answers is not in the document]
JavierArroyoBastida: Sorry I got confused. For a moment I thought that `test_all` was only running python2. And indeed, making python3 the primary script sounds like a good idea to me.

testcase.py
dhblum: Ok, will use version.txt.

.travis.yml
dhblum: Ok per @JavierArroyoBastida comment below and discussions with @SenHuang19 and @kbenne, will run ``test_all`` in python 3 environment, requiring numeric updates to reference results as described above, and run example controllers in python only with python 2 to maintain compatibility for now.  Basically switch what is in the PR right now.

.travis.yml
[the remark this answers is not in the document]
dhblum: Ok great, thanks.  Will use python 3 for ``test_all`` per the comment [here](https://github.com/ibpsa/project1-boptest/pull/336#discussion_r660894926).

testcase.py
dhblum: Done in https://github.com/ibpsa/project1-boptest/pull/336/commits/8067815a5d588496c2fae80676142dfc74e3a769.

.travis.yml
dhblum: Done in https://github.com/ibpsa/project1-boptest/pull/336/commits/2a4af129e48fc9c990cf86ed54bf939fa2a67d10.

.travis.yml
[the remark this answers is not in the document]
dhblum: Done in https://github.com/ibpsa/project1-boptest/pull/336/commits/2a4af129e48fc9c990cf86ed54bf939fa2a67d10.

dhblum:
Also addresses #228 via https://github.com/ibpsa/project1-boptest/pull/339.

dhblum:
This PR also addresses #340 with https://github.com/ibpsa/project1-boptest/pull/336/commits/04a963f0d4f58b3c3c30a17688ffec9496332251.

dhblum:
@SenHuang19 @JavierArroyoBastida @kbenne Looks like tests are passing on this.  Let me know if you have any additional comments or not.

JavierArroyoBastida:
I have no further comments. If I understand correctly, the version upon merging is what we'll call version `0.1.0 `, right? should we label it already? or do we want to wait a bit longer to check the version stability?

dhblum:
@JavierArroyoBastida Yes I think after this is merged we can tag 0.1.0.  I'll initialize a ``changelog.md`` to be able to start tracking changes in future versions and update the version number in version.txt before doing so.

@SenHuang19 I agree, it could make sense to include those values in the config.json just like the other default values.  I could make this change.  And keep the values at 0 for now?

kbenne:
With respect to versioning, I think all of these improvements related to tracking the changes of the boptest software are very positive. 

I also think though, that as the software settles down, test case versioning will become a thing. It could be that any change to a test case prompts a top level project version change, but I'm not sure that is what you want. You might want the test cases themselves to be versioned, and that might be what folks care more or equally about.

dhblum:
@kbenne I agree test case versioning is important.  Perhaps it will make sense down the line to version test cases separately from the rest of software.  In fact, this is how I envisioned it upon writing the proposal for our current funding.  However, I think what could matter in the end is whether the results should be expected to be the same or not.  And this may change not only due to test case updates, but software updates as well - maybe based on updated versions of packages or ways of doing things as features are enhanced and maintained.  Therefore, at this point I'd lean towards starting with one versioning system at the top level that includes the software and test cases.  Minor bumps in versions could be from changes that don't affect results, while more major bumps in version maybe do affect results, add new test cases, or change functionality in a non-backwards compatible way.  The changelog could also track changes to test cases and software separately just to be more clear.

dhblum:
Hm, but what if we add new features, but results don't change.  We may still want to somehow indicate that results from two different versions are comparable...

dhblum:
@SenHuang19 

> should we allow individual testcase to have its "default" start_time and "default" warmup_period? We could add those default values in the config.json

Done in https://github.com/ibpsa/project1-boptest/pull/336/commits/a442cf12188ef3c3b83aa1e2e5573fa99b799192.

testing/utilities.py
```diff
@@ -569,7 +569,7 @@ def test_advance_false_overwrite(self):
         elif self.name == 'bestest_hydronic_heat_pump':
             u = {'oveTSetHea_activate':0, 'oveTSetHea_u':273.15+22}
         elif self.name == 'multizone_residential_hydronic':
-            u = {'conHeaSal_oveTsetHea_activate':0, 'conHeaSal_oveTsetHea_u':273.15+22,
+            u = {'conHeaRo1_oveTSetHea_activate':0, 'conHeaRo1_oveTSetHea_u':273.15+22,
                  'conPumHea_oveActHea_activate':0, 'conPumHea_oveActHea_u':1}
```
kbenne, line 573: I think the names on this line might be out of date too.

dhblum: Fixed in https://github.com/ibpsa/project1-boptest/pull/336/commits/969daad367cdbe8e0496e439ecdf523c287b52a0.

dhblum: Now fixed in https://github.com/ibpsa/project1-boptest/pull/336/commits/977b0c1ba4806f5fcec55cd087b712b3b9687f92.

dhblum:
@SenHuang19 @JavierArroyoBastida @kbenne Let me know if you have any other comments or approve merging.

JavierArroyoBastida (approved):

JavierArroyoBastida:
I think this is ready for merging. @dhblum regarding your comment:

> Hm, but what if we add new features, but results don't change. We may still want to somehow indicate that results from two different versions are comparable...

I'd use the convention that API changes, new emulators and any changes affecting results would bump the second decimal, and all other changes would bump only the third. Even though we could see some new features as major changes from the development perspective, if they don't affect results, they may be seen as minor from the users' perspective.

SenHuang19:
I believe all the issues we discussed have been resolved.
Just last comment: do we need to push the docker image to docker hub?

dhblum:
@SenHuang19 I don't think we need to.  As is architected now, the image is created for a specific test case using the command ``make build TESTCASE=<testcase>`` as indicated on the readme, which I don't think takes too long.  I suggest we keep this process for the current version.

PR #339, opened 2021-06-30
Issue228 Config to Json and in FMU

dhblum:
This is a proposal for #228 and also compiling the ``config`` file into the FMU.  It is something that will make boptest-service easier to integrate after some discussion with @kbenne and I think in general it also makes more sense so that the test case fmu has all of the information needed to run test, rather than having a separate ``config.py`` that is outside the fmu.  In particular, this PR:

- Changes the ``config.py`` in each test case to ``config.json`` and moves it to within a test cases's ``/models`` directory.
- Updates ``data/data_manager.py`` as needed to compile the ``config.json`` within the test case fmu (uses a cherry-picked commit 83c2b73bfeade1415fa492da4d53ef0f20a1dcbb and then edits for use with json) and load within ``testcase.py``.
- Updates ``testcase.py`` to use the data from the loaded ``config.json`` instead of from importing ``config.py``.  
- Updates Design Requirements and Guide documentation to include the ``config.json``.

@kbenne Let me know what you think of this.  I know we said we could keep as ``config.py`` but on second thought I have concerns about diverging the boptest-service branch in this regard and if we're going to make changes to ``config.py``, might be best to do it all at once.  @JavierArroyoBastida it'd be good to get your thoughts as well.

Note it's a branch from ``issue335_verRel``.


dhblum:
Tests fail here in the python2 because wrapped fmus need to be updated.  Would do so after https://github.com/ibpsa/project1-boptest/pull/338 is merged.

dhblum:
A to-do is to make the ``fmupath`` a parameter to the constructor of ``testcase.py``.

JavierArroyoBastida:
I think it makes sense to have the config file as a json and wrapped within the fmu. I've also had a look and the implementation so far and is looking good to me. Thanks for moving this forward!

Issue #351, opened 2021-08-11
Bestest Hydronic HP images in Chrome


JavierArroyoBastida:
The images used for documenting the bestest_hydronic_heat_pump case are not visible in `doc/index.html` when using Chrome. 
Removing the last `\` in the `src` argument for their import has proven to resolve the problem. This issue is to implement this edit.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/352.

PR #352, opened 2021-08-11
Remove \ when importing images.

JavierArroyoBastida:
For https://github.com/ibpsa/project1-boptest/issues/351.


dhblum:
Thank you!  Thoughts on suggested update to release notes?  I think it would be good to keep track of changes like this.

Issue #293, opened 2021-02-23
TestCase: FRP


dhblum:
This issue is to add the FRP test case to the repo.

dhblum:
Linking https://github.com/ibpsa/project1-boptest/pull/679.

PR #357, opened 2021-09-10
initial model

ybae0330:
FRP initial model to be reviewed.


dhblum:
Thanks @ybae0330!  Currently, you have your files in a new directory in the repo, ``FRPtestcase/BOPTEST``.  Can you please instead move them to a directory within ``testcases``, similar to how they are set up for other test cases?  That is, create a new directory within ``testcases`` with a brief, descriptive name, then within that, create ``doc`` and ``models``, and put your modelica package within a directory in ``models``.

ybae0330:
@dhblum Just updated the location of the model. A new directory is created within testcases. Thanks!

dhblum:
Note is for #293

Issue #360, opened 2021-09-24
Add tagging feature


kbenne:
This is an enhancement request to implement a strategy for associating tags with control points, and for returning tag data via the BOPTEST API. This should be implemented at the FMU level. An additional enhancement should be developed for capturing tags at the Modelica level.

PR #361, opened 2021-09-24
Include tags in the /measurements and /inputs APIs

kbenne:
ref #360

### DIFF
--- a/testcase.py
+++ b/testcase.py
@@ -645,6 +645,7 @@ class TestCase(object):
                     maxi = None
             var_metadata[var] = {'Unit':unit,
                                  'Description':description,
+                                 'tags': self.tags_json.get(var),
                                  'Minimum':mini,
                                  'Maximum':maxi}
 
--- a/testcases/testcase1/models/tags.json
+++ b/testcases/testcase1/models/tags.json
@@ -1,0 +1,22 @@
+{
+  "oveAct_u": {
+    "writable": "m:",
+    "unit": "s:W"
+  },
+  "PHea_y": {
+    "point": "m:",
+    "sensor": "m:",
+    "unit": "s:W"
+  },
+  "TRooAir_y": {
+    "point": "m:",
+    "sensor": "m:",
+    "temp": "m:",
+    "unit": "s:C"
+  },
+  "CO2RooAir_y": {
+    "point": "m:",
+    "sensor": "m:",
+    "unit": "s:ppm"
+  }
+}
--- a/data/data_manager.py
+++ b/data/data_manager.py
@@ -215,6 +215,8 @@ class Data_Manager(object):
         self.days_path = os.path.join(models_dir, 'days.json')
         # Find the config.json path
         self.config_path = os.path.join(models_dir, 'config.json')
+        # Find the tags.json path
+        self.tags_path = os.path.join(models_dir, 'tags.json')
 
         if os.path.exists(resources_dir):
             # Find all files within Resources folder
@@ -251,6 +253,13 @@ class Data_Manager(object):
                              os.path.join('resources', 'config.json'))
         else:
             warnings.warn('No config.json found for this test case')
+
+        # Write a copy of tags.json to the fmu resources folder
+        if os.path.exists(self.tags_path):
+            self.z_fmu.write(self.tags_path,
+                             os.path.join('resources', 'tags.json'))
+        else:
+            warnings.warn('No tags.json found for this test case')
 
         # Close the fmu
         self.z_fmu.close()
@@ -384,6 +393,14 @@ class Data_Manager(object):
         # Load config json
         json_str = z_fmu.open('resources/config.json').read()
         self.case.config_json = json.loads(json_str)
+        # Load tags json
+        # This is currently not a required file
+        try:
+            json_str = z_fmu.open('resources/tags.json').read()
+        except:
+            # If there is no tags file then create an empty dictionary
+            json_str = '{}'
+        self.case.tags_json = json.loads(json_str)
 
         # Find the test case data files
         files = []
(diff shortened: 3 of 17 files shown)


dhblum:
Thanks @kbenne for this.  I like having the tags.json and it's similarities to other .json data types we have, in terms of its inclusion in the test case fmu and then expression through the API.  I have a couple of comments/questions though:
1. While we had discussed leveraging the ``/measurements`` and ``/inputs`` api endpoints, and these indeed indicate how one could pass this information through to the user, one concern I'm realizing is that it then focuses on tags for I/O, but does not allow for (at least intuitively) tags for things like pieces of equipment, which are present in e.g. Haystack.  I'm wondering if a new api endpoint then is needed to allow for this, and the return would be a tag report?  Either something dedicated to equipment, or I'm thinking maybe more generally for all tags may be better.
2. The example ``tags.json`` is I think based on Haystack.  Just curious, what does ``m:`` mean?

kbenne:
1. Indeed, BOPTEST API doesn't really expose the intermediate layers. To me, BOPTEST is just exposing the leaves of the tree. I'm not sure what the answer is. One thing that might be part of the solution is to eventually just support Haystack API to compliment the BOPTEST API. The Haystack API offering the full view of the model, the BOPTEST API sticking to I/O.  In my mind they don't have to be exclusive, they can be complimentary and exist and be used simultaneuously. We could alternatively add endpoints to BOPTEST, but does that eventually start to just mirror Haystack, but less good?

2. That one is easy. Yes the tags file is following haystack, and  m: is a marker tag. https://project-haystack.org/doc/docHaystack/Kinds#marker Although it appears that Haystack is saying that json should use the full string "marker". I think they must have changed that in the last year or so.

testcase.py
```diff
@@ -645,6 +645,7 @@ def _get_var_metadata(self, fmu, var_list, inputs=False):
                     maxi = None
             var_metadata[var] = {'Unit':unit,
                                  'Description':description,
+                                 'tags': self.tags_json.get(var),
```
dhblum, line 648: Does it make sense to capitalize ``'tags'`` like other meta data?

testcases/testcase1/models/tags.json
```diff
@@ -0,0 +1,22 @@
...
+  "PHea_y": {
+    "point": "m:",
+    "sensor": "m:",
+    "unit": "s:W"
+  },
+  "TRooAir_y": {
+    "point": "m:",
+    "sensor": "m:",
+    "temp": "m:",
+    "unit": "s:C"
```
dhblum, line 15: This unit should be K.

dhblum (changes requested):
@kbenne I think eventually it would be good to at least be able to specify (in a test case) and get (as a user) a full haystack model, which if I understand correctly could be implemented as a json of tags.  Does it require a full haystack API implementation to get the haystack model, or does one without the other sound ridiculous?  A haystack API could indeed live alongside and be complementary to the current BOPTEST API, but with notable overlaps in functionality being ``PointWrite``-``/advance`` (though advance also advances the simulation) and ``HisRead``-``/results``.

In the meantime for this PR, can you :
- Address my inline comments and
- Update design documentation to include ``tags.json`` in the [test case directory spec](https://github.com/ibpsa/project1-boptest/blob/master/docs/DesReqGui/source/testcasedev.rst#test-case-repository) (comment that it is optional right now), and also add a section and spec for ``tags.json`` as you've implemented similar to [other json specs](https://github.com/ibpsa/project1-boptest/blob/master/docs/DesReqGui/source/testcasedev.rst#time-period-specification-and-json-mapping).
- Update [release notes](https://github.com/ibpsa/project1-boptest/blob/master/releasenotes.md) with this edit

testcase.py
kbenne: certainly. good catch.

testcases/testcase1/models/tags.json
kbenne: > @kbenne I think eventually it would be good to at least be able to specify (in a test case) and get (as a user) a full haystack model, which if I understand correctly could be implemented as a json of tags. Does it require a full haystack API implementation to get the haystack model, or does one without the other sound ridiculous? A haystack API could indeed live alongside and be complementary to the current BOPTEST API, but with notable overlaps in functionality being `PointWrite`-`/advance` (though advance also advances the simulation) and `HisRead`-`/results`.
> 
> In the meantime for this PR, can you :
> 
> * Address my inline comments and
> * Update design documentation to include `tags.json` in the [test case directory spec](https://github.com/ibpsa/project1-boptest/blob/master/docs/DesReqGui/source/testcasedev.rst#test-case-repository) (comment that it is optional right now), and also add a section and spec for `tags.json` as you've implemented similar to [other json specs](https://github.com/ibpsa/project1-boptest/blob/master/docs/DesReqGui/source/testcasedev.rst#time-period-specification-and-json-mapping).
> * Update [release notes](https://github.com/ibpsa/project1-boptest/blob/master/releasenotes.md) with this edit

As far as I know, you don't have to implement all of the Haystack API. In fact the API specifies an `/ops` endpoint that returns a form of documentation for what endpoints the server is supporting. I think `read` and `nav` would be sufficient for model discovery.

dhblum: Then the use of ``/ops``, ``read``, and ``nav``, along with implementing an example "full" model, I think could be a good next goal, at least in the direction of haystack support.

dhblum: For https://github.com/ibpsa/project1-boptest/pull/361#discussion_r718007783 the unit has been changed in https://github.com/ibpsa/project1-boptest/pull/361/commits/0868a08aab0dc14fd93ddd9d26c21c773dd2ab9c.

dhblum:
Unit tests pass at commit https://github.com/ibpsa/project1-boptest/pull/361/commits/262c9f549156f61a4067907ee0cdee41c83f1a20.

Further commits after this store Haystack tags in the ``bestest_air`` case and prototype a workflow for their usage in creating the ``tags.json``.  In particular, they:
- Update the ``bestest_air`` model to include signal exchange blocks that use Haystack tags as implemented in a fork of the Modelica Buildings Library at https://github.com/dhblum/modelica-buildings/tree/issue_boptest_360_haystack_tags.  Such updates to Modelica Buildings Library via Modelica IBPSA Library still needs to be done, however, changes will be implemented in development branches of those libraries that use MSL 4.0, which BOPTEST needs to resolve through use of compilation tools other than JModelica (e.g. OpenModelica and Spawn).
- Update the ``parser.py`` to parse the Haystack tags from the signal exchange blocks and create the ``tags.json``.
- Update the ``bestest_air`` wrapped fmu as an example of making the Haystack tags available through BOPTEST.
- Will not pass unit tests because the utilized Modelica Buildings Library for compilation of test cases during tests needs to be changed to the forked implementation and/or the ``parser.py`` in this branch needs to be updated to compile cases that do not use the implementation of Haystack tags in the signal exchange blocks.

The output of ``/measurements`` using the ``bestest_air`` test case in this branch now gives for example:

Screen Shot 2022-04-06 at 5 25 12 PM

Issue #362, opened 2021-09-29
Update test cases to Buildings v8.0.0 and IDEAS v2.2.1


dhblum:
This issue is to update test case FMUs and ``library_versions.json`` to use recently released versions of Modelica libraries, namely Buildings v8.0.0 and IDEAS v2.2.1.  Also to remove any dependencies on IBPSA Modelica library itself, since it is merged into both Buildings and IDEAS.

PR #363, opened 2021-09-30
Issue362 update modelica libraries

dhblum:
This is for #362.  Still to be done:

- [x] Remove IBPSA Modelica dependency from development test cases
- [x] Update release notes
- [x] Update all FMUs to be sure


dhblum:
@JavierArroyoBastida Can you review this?  Note that this update of Modelica versions does not change unit test reference results for kpis or results for any test case except for testcase2 since it uses an updated controller model in Buildings v8.0.0.  But this is no problem since testcase2 is for development purposes only.

JavierArroyoBastida (commented):
This is looking good. Thanks @dhblum for taking care of this! 
Just out of curiosity, what's the reason for the updated controller of testcase2?
To clarify, this PR slightly updates reference results also for testcases other than testcase2, although it is true that these changes do not significantly change KPIs and only affect the initial values because of the updated weather data files. E.g. for the `bestest_hydronic` case:
image

image

wrapped models still need to be updated, right?

dhblum:
@JavierArroyoBastida The updated controller in testcase2 is because it is based on a single zone VAV controller in the Buildings library that's been updated since the original implementation of testcase2.  Namely this one: https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Air/Systems/SingleZone/VAV/ChillerDXHeatingEconomizer.mo.

Yes there are slight changes due to slightly changed weather files.

If the PR looks good, yes all that's needed before merging is updating fmus.  Wanted to wait until approved otherwise.

JavierArroyoBastida:
OK I see, thanks for the note on the updated controller! This looks good to me and I think fmus can be updated already.

Issue #365, opened 2021-10-06
Update Dockerfile commands


dhblum:
This issue is to add the use of ``CMD python restapi.py && bash`` and ``EXPOSE 5000`` in ``Dockerfile`` to encode this default behavior there instead of the ``makefile``.  Use of ``docker run -P`` will publish the container port 5000 (on which the api is listening) to a random host port, while continued use of ``docker run -p 127.0.0.1:5000:5000`` is needed to publish specifically to port 127.0.0.1:5000 of the host.  See https://docs.docker.com/engine/reference/builder/#expose.

dhblum:
As described in https://github.com/ibpsa/project1-boptest/pull/366, this issue also replaces the use of the makefile with docker-compose, which is good especially for Windows users to alleviate needing to install make and makes it easier to use native docker commands. The makefile is kept, however and updated as needed to make ``make build`` and ``make run`` deprecated.

Note https://github.com/ibpsa/project1-boptest/pull/366 closed this issue.

PR #366, opened 2021-10-06
Issue365 docker commands

dhblum:
This is for #365, by adding ``CMD`` and ``EXPOSE`` to ``Dockerfile``.  It also updates ``makefile`` to not have to run the python command to start the API for both ``run`` and ``run-detached`` targets.  Some local tests indicate this is backwards compatible with how the boptest container is built and deployed now, subject to unit test results, but enables some capabilities if users work with docker directly.

@kbenne Would this impact service at all?
@mschaar Is this what you were suggesting would help?

UPDATE: This now also replaces the use of the ``makefile`` with ``docker-compose``, which may be good especially for Windows users to alleviate needing to install make.  Now, boptest test cases would be able to be deployed and shutdown with ``docker-compose`` commands.  ``makefile`` kept and updated as needed to make ``make build`` and ``make run`` deprecated.


JavierArroyoBastida:
I have successfully implemented your option 1 in Windows with PowerShell, although I used a slightly modified version of it. Particularly, I've run `setx TESTCASE "testcase1"` and then, in a separate command, I've run `docker-compose up`, which has built and deployed the testcase. I have tried to run both in the same command but I get: `The token '&&' is not a valid statement separator in this version.`
when trying to run the unit tests for testcase1 I get the following error:
```
cd .. && TESTCASE=testcase1 docker-compose up -d
'TESTCASE' is not recognized as an internal or external command,
operable program or batch file.
make: *** [test_testcase1] Error 1
```

Again, this is related to how environment variables are set differently in Linux and Windows.

SenHuang19:
Hi Javier,

I used "cmd /C "set TESTCASE=testcase1 && make build"" in Windows with
PowerShell and it works.

JavierArroyoBastida:
Thanks a lot @SenHuang19 for the hint! I guess you mean: `cmd /C "set TESTCASE=testcase1 && docker-compose up"`? since the aim is to avoid the use of `make` here, right? In my case, when I try `cmd /C "set TESTCASE=testcase1 && docker-compose up"` My only output log in PowerShell is:
```
ERROR: no such image: boptest_testcase1 : invalid reference format
```

I've also tried with `make: *** No rule to make target `build'.  Stop.` and in this case the output is:

```
make: *** No rule to make target `build'.  Stop.
```

Note that I'm checking out commit: https://github.com/ibpsa/project1-boptest/pull/366/commits/151f2cf665bfbb1cc4b42457785227722b325a21

SenHuang19:
> Thanks a lot @SenHuang19 for the hint! I guess you mean: `cmd /C "set TESTCASE=testcase1 && docker-compose up"`? since the aim is to avoid the use of `make` here, right? In my case, when I try `cmd /C "set TESTCASE=testcase1 && docker-compose up"` My only output log in PowerShell is:
> 
> ```
> ERROR: no such image: boptest_testcase1 : invalid reference format
> ```
> 
> I've also tried with `make: *** No rule to make target `build'. Stop.` and in this case the output is:
> 
> ```
> make: *** No rule to make target `build'.  Stop.
> ```
> 
> Note that I'm checking out commit: [151f2cf](https://github.com/ibpsa/project1-boptest/commit/151f2cf665bfbb1cc4b42457785227722b325a21)

Hi Javier, the command I mentioned is just to show how to define the environment variable when running a command in the PowerShell.

The first error you have seems to be missing the image. Did you try to build before up?
Fr the second error, I was not aware that we are replacing "make" with "docker-compose". Sorry for the confusions

dhblum:
Thanks @JavierArroyoBastida and @SenHuang19 for checking on this!  One comment is that ``docker-compose up`` should build the image first if it is not already built, so a separate ``docker-compose build`` command should not be needed.

dhblum:
@JavierArroyoBastida I'm not sure why if ``setx TESTCASE "testcase1"`` and ``docker-compose up`` work in separate commands, then why they won't work with ``&&`` between them in a single command?  Note it looks like when you tried with ``&&`` you used ``set`` instead of ``setx``.  Not sure the difference of using the ``cmd /C`` or not as suggested by @SenHuang19, I'm not a Windows user.

Also, can you try with cygwin if in the unit test make file you can use something like ``export TESTCASE=testcase1 && docker-compose up``?

Note that my last commit builds a base docker image first (if needed) and then simply loads the correct fmu into the container upon deployment according to the chosen test case.  Much faster deployment of a new test case if the base image is already built.

JavierArroyoBastida:
Apparently, the reason why I couldn't deploy the test case in one line is that the `&&` operator does not work in Windows PowerShell. See [this](https://stackoverflow.com/questions/65627536/the-token-is-not-a-valid-statement-separator-in-this-version) or [this](https://stackoverflow.com/questions/563600/can-i-get-or-and-to-work-in-powershell) explanation. 
I've used the `(setx TESTCASE "testcase1") -and (docker-compose up)` command in PowerShell as suggested in the links above and it works, although the output log from the commands is now shown in the console. I only get as an output:
```
Creating boptest_boptest_1 ... done                                                                                                   
```
Something that surprises me is that the name of the image is `boptest_boptest_1` and not `boptest_testcase1`. Is that the intended behavior? is this some default naming convention of docker-compose?

dhblum:
@JavierArroyoBastida Some recent updates I think have addressed your comments, for instance on the name of the image that's created.  Any chance you can take another look at this and let me know if it's ok to merge?

JavierArroyoBastida:
Thanks @dhblum for the updates! I've gone through the edits and they look good to me. Note that I couldn't test with my Windows PC right now because I changed my laptop and don't have the old one with me right now. I could do that if you think it's needed but only later this week.

JavierArroyoBastida:
OK, I'll do!

Issue #378, opened 2021-10-29
Add related publications list


dhblum:
This issue is to add a related publications list.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/383.

PR #379, opened 2021-10-29
Patch - add journal 2021

dhblum:
This is to add the recently published journal article to the readme.  I suggest also in https://github.com/ibpsa/project1-boptest/issues/378 to add other related publications.


JavierArroyoBastida (approved):
Looking good!

JavierArroyoBastida:
Are we closing https://github.com/ibpsa/project1-boptest/issues/378 with this PR? or shall we use the same PR to add related publications?

dhblum:
I think don't close #378 with this PR but use another PR to add more related publications.

Issue #381, opened 2021-12-01
Fix find mixed day scenario


dhblum:
The algorithm for finding the mixed day scenario in https://github.com/ibpsa/project1-boptest/blob/master/data/find_days.py#L122 uses the max daily heating and cooling loads from a given day, rather then the daily sum of heating and cooling loads.  This issue will correct this and update the mixed day scenario days for the ``bestest_air`` case.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/382.

PR #382, opened 2021-12-06
Issue381 find days

dhblum:
This is for #381 and updates ``bestest_air`` accordingly.


JavierArroyoBastida (approved):
Thanks @dhblum for realizing this and implementing the correction. I have no further suggestions to make for this PR, so I believe it can be merged as it is. Now the mix day is calculated as it was originally proposed in [here](https://github.com/ibpsa/project1-boptest/issues/270#issuecomment-774719263) and also corresponds to the definition in [our paper](https://www.tandfonline.com/doi/full/10.1080/19401493.2021.1986574).

dhblum:
Thanks @JavierArroyoBastida for the review!  I merged master and resolved conflicts (release notes).  I also indicated this on release notes now as a backwards-compatible change (since the API and everything didn't change) but does change benchmark results.  Let me know if you disagree, otherwise, I will merge once tests pass a final time.

JavierArroyoBastida:
@dhblum I think compatibility should refer to usability rather that benchmarking results, plus it is indicated in the `releasenotes.md` file that those results do change with the edit. So yes, I agree.

Issue #384, opened 2021-12-13
Multizone residential hydronic simulation improvements


dhblum:
When simulating the ``multizone_residential_hydronic`` test case with overwriting heating and cooling set points to 20 and 20.5 with a /step of 240 s, the simulation stalls at 4.644000E+06.  The following two updates alleviate this problem such that the simulation completes for the full year with this configuration, also for the baseline implementation, and does not significantly change KPIs:
1. In ``MultiZoneResidentialHydronic.Building.Equipement.Heating.Boiler`` change one-port temperature sensors to two-port sensors
2. In ``MultiZoneResidentialHydronic.TestCase`` switch out ``Buildings.Airflow.Multizone.DoorDiscretizedOperable`` for ``Buildings.Airflow.Multizone.DoorDiscretizedOpen`` since input signals for doors are always 1.0 (open).

Example KPI changes (peak heat day constant prices):
Current, Proposed
tdis_tot | 22.0030645202424, 22.00312716332
idis_tot | 9114.58979568513, 9114.59953485554
ener_tot | 8.14359225757814, 8.1423625182175
cost_tot | 0.793424710894034, 0.79331256851389
emis_tot | 1.42593875005342, 1.42571617574018
time_rat | 0.000110403521596, 0.000102698924987

PR #385, opened 2021-12-13
Issue384 multizone residential hydronic simulation improvement

dhblum:
This is for #384.  Still need to:
- [x] update wrapped.fmu


dhblum:
Ready to merge once tests pass.

Issue #396, opened 2022-01-24
TestCase: Single Zone Commercial Air


dhblum:
This issue is to create the single zone commercial air test case as specified in IBPSA Project 1 WP1.2.

It will create a test case based on a single store from the [DOE prototype building model "strip mall"](https://www.energycodes.gov/prototype-building-models) based in the New York climate (mixed).  This will be a single zone (LG_STORE_2) with packaged rooftop unit (RTU) with DX and natural gas furnace.  

Model development on issue396 development branch.  ToDos (continually updated):

- [x] Update multistage dx coil performance map
- [x] Update supply fan performance map
- [x] Update natural gas furnace performance
- [ ] Finalize equipment sizing
- [x] Update controls for staged cooling, heating, and fan based on eplus spec
- [x] Add controls for economizer
- [x] Add signal exchange blocks
- [x] Add CO2 tracking
- [x] Add data csv files
- [x] Add config.json
- [ ] Simulate wrapped.fmu for 1 year in boptest
- [ ] Update days.json
- [ ] Add documentation
- [ ] Add unit tests
- [ ] IBPSA WP1.2 Review

PR #397, opened 2022-01-24
Issue 396 initial implementation

dhblum:
Initial implementation of single zone commercial air model.


dhblum:
This was for #396

PR #405, opened 2022-02-17
Fix issue using parser with python3

### LINKED ISSUE #404: Parser is not compatible with python 3
There are very minor issues preventing the test case parser from working with both python 2 and 3.

kbenne:
close #404

### DIFF
--- a/parsing/parser.py
+++ b/parsing/parser.py
@@ -49,8 +49,8 @@ def parse_instances(model_path, file_name):
     if fmu.get_version() != '2.0':
         raise ValueError('FMU version must be 2.0')
     # Get all parameters
-    allvars =   fmu.get_model_variables(variability = 0).keys() + \
-                fmu.get_model_variables(variability = 1).keys()
+    allvars =   list(fmu.get_model_variables(variability = 0).keys()) + \
+                list(fmu.get_model_variables(variability = 1).keys())
     # Initialize dictionaries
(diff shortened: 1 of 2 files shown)


dhblum:
Thanks @kbenne for the PR.  Is a change not also needed at https://github.com/ibpsa/project1-boptest/blob/master/parsing/parser.py#L149 and https://github.com/ibpsa/project1-boptest/blob/master/parsing/parser.py#L163?

dhblum:
@kbenne Also, please update the ``releasenotes.md`` in the backwards compatible section with a brief description and issue reference.

kbenne:
> Thanks @kbenne for the PR. Is a change not also needed at https://github.com/ibpsa/project1-boptest/blob/master/parsing/parser.py#L149 and https://github.com/ibpsa/project1-boptest/blob/master/parsing/parser.py#L163?

I believe, based on my scenario, the answer is no. In python 3, the result of `keys()` can still be iterated, it appears you just can't concatenate them with the `+` operator because they are now ordered lists.

kbenne:
Ah. Thank you. Didn't know the convention of the release notes.

dhblum:
Ok I see, great.  Then this looks good.  I just moved the release notes description to the top (since the top has been most recent changes).  Will merge once tests pass.

dhblum (approved):

Issue #419, opened 2022-04-27
Selecting testcase with docker compose in windows


haraldwalnum:
I have struggled quite a bit with selecting testscases when using docker-compose in windows. Originally, I was not able to use the `setx` command to select testcases. I selected whatever was in the .env file. 
As I understand, the `setx` command sets the env. variable at **user level** :https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/setx
I think the .env file overwrites this.

At a later stage I had trouble setting the testcase also in the .env file. It turned out that (for some reason I dont't know) the evn. variable TESTCASE had been set at shell level. This overwrites what in the .env file https://docs.docker.com/compose/environment-variables/
My solution was to use:
 `> ($env:TESTCASE="<testcase_directory") -and (docker-compose up)`
instead of:
 `> (setx TESTCASE "<testcase_dir_name>") -and (docker-compose up)`

`$env` sets the env. variable at shell level for the current session only.

To me this seems to be a safer approach.
@JavierArroyoBastida I know you have tested a bit with Windows and PowerShell. Have you experienced something similar?

JavierArroyoBastida:
@haraldwalnum pity that it's proving so challenging. I have no extensive experience with `docker-compose` in Windows since I recently changed laptop. All experience I have is reported [here](https://github.com/ibpsa/project1-boptest/pull/366), which is the issue that led to using the command that did not work for you. That is:
```
> (setx TESTCASE "<testcase_dir_name>") -and (docker-compose up)
``` 

@dhblum it's maybe worth editing the `README` for substituting the command with the one suggested by @haraldwalnum, or maybe keeping both indicating that the second one can be used if there is any trouble.

dhblum:
Thanks @haraldwalnum for reporting and also proposing the solution.  A similar issue was reported to me offline and I'm seeing if this solution can help that too.  Can you also just verify your proposed solution doesn't require a ``>`` after ``testcase_directory``?

dhblum:
Confirming the fix works for the offline case I got as well.

haraldwalnum:
> Thanks @haraldwalnum for reporting and also proposing the solution. A similar issue was reported to me offline and I'm seeing if this solution can help that too. Can you also just verify your proposed solution doesn't require a `>` after `testcase_directory`?

the missing `>` is a typo from my side. Should be:
`> ($env:TESTCASE="<testcase_directory>") -and (docker-compose up)` or as an example:
[...]

PR #420, opened 2022-04-27
Issue 419 Setting Test Case in Windows

dhblum:
This is for #419.


dhblum:
@haraldwalnum can you review and approve?

haraldwalnum:
Looks good,  but I would add the missing `>` as commented in the issue

Issue #423, opened 2022-05-13
Add file exclusion to data manager load_data_and_jsons()


dhblum:
The load_data_and_jsons() method of the data manager loads, checks, and tries to process every file within the FMU resources directory with a .csv ending.  This issue is to add the ability to exclude certain files that are needed for FMU simulation but not for access by BOPTEST infrastructure.  One example is for Spawn simulation, the inclusion of schedules as csv files.  Note a current workaround is including these files as .txt.

Relevant line: https://github.com/ibpsa/project1-boptest/blob/c464c089061b1659ed4d477e8c0f85e30b23a4a8/data/data_manager.py#L391

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/426.

PR #426, opened 2022-05-24
Issue423 data exclusion list

dhblum:
This is for #423.  Adds optional resource file exclusion list to test case ``config.json``, filter in ``data_manager.py``, and updates documentation.


dhblum:
@kbenne Confirm this would address your issue for Spawn schedules?  @JavierArroyoBastida FYI, let me know if you have issues with merging.

PR #436, opened 2022-06-14
Issue73 messaging

SenHuang19:


testcase.py
```diff
@@ -644,19 +675,19 @@ def get_kpis(self):
...
+    def set_forecast_parameters(self, horizon, interval):
         '''Sets the forecast horizon and interval, both in seconds.
 
         Parameters
         ----------
-        horizon: int
+        horizon: float
             Forecast horizon in seconds.
-        interval: int
+        interval: float
```
dhblum, line 690: @SenHuang19 A reason to change interval and horizon to float instead of int?  I think it may be safer to use int?

testcase.py
```diff
@@ -312,11 +318,11 @@ def initialize(self, start_time, warmup_period, end_time=np.inf):
 
         Parameters
         ----------
-        start_time: int
+        start_time: float
             Start time of simulation to initialize to in seconds.
-        warmup_period: int
+        warmup_period: float
```
dhblum, line 323: @SenHuang19 A reason to change warmup_period and start_time to float instead of int?  I think it may be safer to use int?

testcase.py
```diff
@@ -441,21 +462,25 @@ def set_step(self, step):
...
         payload = None
-        try:
-            step = float(step)
-        except:
+        if not isinstance(step, int):
             payload = None
             status = 400
-            message = "Invalid value for the control step"
-            logging.warning(message)
+            message = "parameter 'step' must be an integral number but is {}".format(type(step))
```
dhblum, line 468: @SenHuang19 Change "integral" to "integer"

testcase.py
```diff
@@ -288,12 +291,15 @@ def advance(self, u):
...
+                else:
+                    message = alert_message               
                 logging.info(message)
                 return status, message, payload
 
             else:
-                # Error in
+                # Errors in the simulation
                 status = 500
                 message = "Failed to advance simulation: {}"
```
dhblum, line 304: @SenHuang19 I think this needs a .format(res) at the end so the full error message is returned, just like it is logged in Line 306?

dhblum (changes requested):
@SenHuang19 Generally this is looking very good, thank you so much.  A few in-line comments though I'm hoping you can address which I've marked up.

testcase.py
```diff
@@ -245,22 +246,24 @@ def advance(self, u):
...
-                            return status, message, None
+                            message = "Invalid value for {} - value must be a float but is {}".format(key, type(u[key]))
+                            logging.error(message)
+                            return status, message, None                            
                         # Check min/max if not activation input
                         if '_activate' not in key:
-                            checked_value = self._check_value_min_max(key, value)
+                            checked_value, message = self._check_value_min_max(key, value)
+                            if message is not None:
+                                alert_message = alert_message +'; '+ message
```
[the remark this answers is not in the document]
dhblum: Also in this case, please log as logger.warning.

SenHuang19:
@dhblum As per our earlier discussion, we added more unit tests and made floats to be acceptable for the time-related parameters. However, the unit test fails for the new commit. I spent some times on this and realized the issue is that
the changes I added to the resapi.py,

    elif '_u' in key:
        parser_advance.add_argument(key, type=float)

yields different results as

    value = float(u[key])

As an example, the former generates results like 

    -11.1297454685654

while the latter generates results like
 
    -11.1297454685655

So we either need to update the reference dataset or come out a better way to handle the decimals.

Let me know your thoughts and we can chat about it in Toronto

dhblum:
Thanks a lot @SenHuang19 for the update.  The difference in the results you show are extremely small numeric differences and not perceivable to humans.  I don't mind updating the reference results to have the implementation we prefer.

I will take a quick look, but my plan is to merge your changes and apply additional updates and clean ups as we discussed.  I'll take care of. updating reference results accordingly.  I'll ping you when you can have a final review and/or if more discussion is needed. Thanks again!

dhblum (approved):
Thanks for the changes.  Will merge as is.

Issue #73, opened 2019-05-14
User messages and logging


dhblum:
BOPTEST needs a mechanism(s) to communicate warnings, errors, and other information with the user.  Perhaps this could be done though logging, API messages, or other UI tools.

SenHuang19:
> @SenHuang19 I merged #272 and made some messaging and other clean up updates in [12cf85d](https://github.com/ibpsa/project1-boptest/commit/12cf85dd1fd64ca5485a0122f5381bba934e5981). Can you make a PR to branch `issue73_messaging` with added unit tests?
> 
> I also think there are improvements (additional checks or check modifications) that could be made on the input checking and error reporting for each `testcase.py` function to be more effective and descriptive, see below. I do not think they are hard to implement now that you've got the general structure in place. I can also help implement these updates in parallel to you implementing unit tests on the existing. Then, we can add any needed additional unit tests to your initial set up. Let me know if you agree or other thoughts.
> 
> **1. General**
> 
> * [ ]  statuses of 400 or 500 should be logged as `logger.error` instead of `logger.warning`.
> 
> **2. set_scenario**
> 
> * [ ]  check "electricity_price" parameter is "constant", "dynamic", or "highly_dynamic"
> * [ ]  check "time_period" parameter is an acceptable string based on those in days.json.
> * [ ]  check if initialization actually throws error
> 
> **3. set_forecast_parameters**
> 
> * [ ]  check horizon and internal separately and return error message specific for each
> * [ ]  check values are floats, rather than that they can be translated to floats
> 
> **4. get_results**
> 
> * [ ]  check start_time and final_time separately and return error message specific for each
> * [ ]  check values are floats, rather than that they can be translated to floats
> * [ ]  check var is valid point name and return error accordingly (check against inputs and measurements list)
> 
> **5. get_measurements**
> 
> * [ ]  use traceback in message like other if status is 500
> 
> **6. get_inputs**
> 
> * [ ]  use traceback in message like other if status is 500
> 
> **7. set_step**
> 
> * [ ]  check step is integer instead of can be translated to float
> * [ ]  check step is non-negative
> 
> **8. get_step**
> 
> * [ ]  structure like get_inputs or other gets and similarly use traceback in message like other if status is 500
> 
> **9. initialize**
> 
> * [ ]  check start_time and warmup_period separately and return appropriate error message
> * [ ]  check start_time and warmup_period are integer instead of can be translated to float
[...]

PR #272, opened 2021-01-12
Issue73 add optional logging and standardize return messages

SenHuang19:

PR #441, opened 2022-06-26
Issue73 messaging

dhblum:
For #73.


dhblum:
@SenHuang19 Tests pass.  Want to take a look and give it another review? I think we're pretty much there.

dhblum:
@JavierArroyoBastida This is a PR for the API changes that introduce more structure and opportunity to give messages and errors.  Take a look and see if you have any comments?

SenHuang19:
@dhblum Very nice work, just a few small comments:

1. https://github.com/ibpsa/project1-boptest/blob/issue73_messaging/testcase.py#L267 to #272  seems to be not necessary.
2. https://github.com/ibpsa/project1-boptest/blob/issue73_messaging/restapi.py#L92: do we need "required=True"?
3. For the javascript and julia code, we could create a function to avoid referring "payload" directly multiple times

dhblum:
@SenHuang19 Thanks for the quick review.  

1. L267 - L272 is necessary to ensure the _activate value is either 1 or 0, and nothing else.
2. Yes, added in https://github.com/ibpsa/project1-boptest/commit/19700b553c7cf97377adfbb275fd9922e6cd9e38.
3. We could and I understand the limit on code replication, but I'm not that proficient in those languages to do that and I think this is pretty straightforward for now for users to see without too many layers of code.  Do you have a strong feeling about it?

examples/python/interface.py
```diff
@@ -109,59 +136,62 @@ def control_test(control_module='', start_time=0, warmup_period=0, length=24*360
...
-        # If custom KPIs, compute them
+        # If custom KPIs are configured, compute the KPIs
         for kpi in custom_kpis:
             kpi.processing_data(y)  # Process data as needed for custom KPI
             custom_kpi_value = kpi.calculation()  # Calculate custom KPI value
             custom_kpi_result[kpi.name].append(round(custom_kpi_value, 2))  # Track custom KPI value
             print('KPI:\t{0}:\t{1}'.format(kpi.name, round(custom_kpi_value, 2)))  # Print custom KPI value
         custom_kpi_result['time'].append(y['time'])  # Track custom KPI calculation time
-        # If controller needs a forecast, get the forecast data and update it with controller
+        # If controller needs a forecast, get the forecast data provide the forecast to the controller
```
JavierArroyoBastida, line 178:
```suggestion
        # If controller needs a forecast, get the forecast data and provide the forecast to the controller
```

restapi.py
```diff
@@ -66,129 +103,138 @@ class Advance(Resource):
...
+        status, message, payload = case.get_measurements()
+        return construct(status, message, payload)
+
 
 class Results(Resource):
     '''Interface to test case result data.'''
 
     def put(self):
-        '''PUT request to receive measurement data.'''
+        '''GET request to receive measurement data.'''
```
JavierArroyoBastida, line 161: Isn't `Results` a `PUT` request?

restapi.py
```diff
@@ -208,4 +254,4 @@ def get(self):
 # --------------------------------------
 
 if __name__ == '__main__':
-    app.run(host='0.0.0.0')
+    app.run(host='0.0.0.0', debug=True)
```
JavierArroyoBastida, line 257: Is it strictly needed to run the app in debug mode?

JavierArroyoBastida (approved):
Looking good to me. Thanks for the efforts! I've just added a comment and a couple of questions.

testcase.py
```diff
@@ -238,13 +251,50 @@ def advance(self,u):
...
+                                if float(u[key]) == 1:
+                                    checked_value = 1
+                                elif  float(u[key]) == 0:
+                                    checked_value = 0
+                                else:
+                                    payload = None
+                                    status = 400
+                                    message = "Invalid value {} for input {}. Value must be a boolean, float, integer, or string able to be converted to a 1 or 0, but is {}.".format(u[key], key, type(u[key]))
+                                    logging.error(message)
+                                    return status, message, payload
```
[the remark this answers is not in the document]
dhblum: Yes, the program does continue in the ``except`` statement and checks to see if the string is ``'True'`` or ``'False'``.  If it is anything else, it returns and logs an error indicating such.  Is your proposal to:

1. Check the input is numeric or string.
2. If numeric, check is equal to 1 or 0.  If not, return and log error.
3. Elif string, check is equal to True or False.  If not, return and log error.
4. Else, return and log error.

In that case, I think then logging.error(message) is used three times?

restapi.py
dhblum: Fixed in https://github.com/ibpsa/project1-boptest/pull/441/commits/9c334df9e01752ca7fe43166f049bbbdea5565dd.

restapi.py
dhblum: Fixed in https://github.com/ibpsa/project1-boptest/pull/441/commits/1e93eb94ff86c40971f0ac86a25e002dc500f419.

dhblum:
Thanks @JavierArroyoBastida.  I've addressed your comments in appropriate commits, thanks for pointing out those oversights.  I've also replied to your final comment which is a little more complicated.  Let's continue the discussion there.

testcase.py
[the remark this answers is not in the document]
SenHuang19: I feel we can do 

>
                            try:
                                if float(u[key]) == 1:
                                    checked_value = 1
                                elif  float(u[key]) == 0:
                                    checked_value = 0
                            except ValueError:
                                if u[key] == 'True':
                                    checked_value = 1
                                elif  u[key] == 'False':
                                    checked_value = 0
                            except:
                                payload = None
                                status = 400
                                message = "Invalid value {} for input {}. Value must be a boolean, float, integer, or string able to be converted to a 1 or 0, but is {}.".format(u[key], key, type(u[key]))
                                logging.error(message)
                                return status, message, payload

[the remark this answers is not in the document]
JavierArroyoBastida: My suggestion would be to do something like:
1. `if` input is numeric `and` 1
2. `Elif` input is numeric `and` 0
3. `Elif` input is string `and` "True"
4. `Elif` input is string `and` "False"
5. `Else`: use logging.error(message)

Notice also that the proposal in [here](https://github.com/ibpsa/project1-boptest/pull/441/files/19700b553c7cf97377adfbb275fd9922e6cd9e38#r914174564) enables to use more specific messages, like: "A string was passed but it was neither `True` or `False`, it was ..."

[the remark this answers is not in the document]
dhblum: Ok thanks.  Will update the message though.

dhblum:
@SenHuang19 and @JavierArroyoBastida.  I added a few more commits since you may have reviewed, but all in all, I think this is ready to merge once tests pass and would like to in the next few days.  Let me know if you disagree.  @SenHuang19 let me know also if you have any more comments on the comment [here](https://github.com/ibpsa/project1-boptest/pull/441#discussion_r914878845).

Issue #403, opened 2022-02-12
Add API endpoint for submitting results to dashboard


dhblum:
This is to implement a python method in testcase.py to submit results to the dashboard, using a web address defined as an environment variable.  This will also then need to be extended in the REST API and boptest-service API.  

@kbenne I couldn't find an issue already existing for this, but let me know if you had already created one and I missed it.

dhblum:
First implementation here: https://github.com/ibpsa/project1-boptest/tree/issue403_submit.  It also includes commits from #73 before that gets merged soon, so this is ready to go once #73 is merged.

To dos:
- [ ] Determine how to unit test
- [ ] Resolve https://github.com/ibpsa/project1-boptest/issues/435
- [ ] Update releasenotes and README docs
- [ ] PR and review

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/444.

PR #444, opened 2022-07-12
Issue403 submit

dhblum:
This is for #403.


dhblum:
@kbenne Can you take a look and let me know what you think?  A couple specific points too:
- I proposed an environment variable for the dashboard server.
- Take a look at:  https://github.com/ibpsa/project1-boptest/pull/444/files#diff-9950be71a09e9a0f127636460c8f794665a9fa71ced221e055170a3fc7ad4b6fR1124.  Is this the best [even good] way to handle?
- As noted in https://github.com/ibpsa/project1-boptest/issues/403 still need to write unit tests somehow without posting to the actual dashboard.

testcase.py
```diff
@@ -1054,6 +1062,100 @@ def get_version(self):
...
+                status = 400
+                message = 'Invalid type for one of the tag inputs. They must be strings, but one is a {0}.'.format(type(tag))
+                logging.error(message)
+                return status, message, None
+        # Specify server address and payload
+        dash_server = os.environ['BOPTEST_DASHBOARD_SERVER']
+        payload = {
+          "results": [
+            {
+              "uid": str(int(random.random()*10000000)),
```
kbenne, line 1124: A real uuid might be better. Like this?

https://docs.python.org/3/library/uuid.html

testcase.py
```diff
@@ -1054,6 +1062,100 @@ def get_version(self):
...
+            return status, message, None
+
+        for tag in tags:
+            if not isinstance(tag, str):
+                status = 400
+                message = 'Invalid type for one of the tag inputs. They must be strings, but one is a {0}.'.format(type(tag))
+                logging.error(message)
+                return status, message, None
+        # Specify server address and payload
+        dash_server = os.environ['BOPTEST_DASHBOARD_SERVER']
```
kbenne, line 1120: This env var seems good to me. It appears you used the same name as service is already using which is nice.

testcase.py
```diff
@@ -1054,6 +1062,100 @@ def get_version(self):
...
+            If 400, there was an input error.
+            If 500, an internal error occurred.
+        message: str
+            Includes detailed debugging information
+        payload: None
+
+        '''
+
+        # Check formal scnenario test completed
+        if not self.scenario_end:
```
kbenne, line 1089: I expect Devetry is doing some validation on the payload and returning error codes, but I see no reason not to do this pre checks.

testcase.py
```diff
@@ -1190,3 +1292,24 @@ def _get_full_current_state(self):
         z.update(self.u)
 
         return z
+
+    def to_camel_case(self, snake_str):
```
kbenne, line 1296: I'm sorry we have to do these conversions.

kbenne (approved):

testcase.py
dhblum: Yes, I took it from service ;-)

testcase.py
dhblum: I'm not sure there's anything stopping someone from starting a scenario, running it halfway, and then submitting results, in which case all the payload data would be "valid", but the KPIs would be non-comparable.  This is to make sure a scenario is finished (i.e. time period) before allowing to send results.  There is some error checking on the dashboard side and I handle that starting at https://github.com/ibpsa/project1-boptest/blob/issue403_submit/testcase.py#L1146.

testcase.py
dhblum: Maybe this gets cleaned up down the line or something.  Fine for now.

testcase.py
dhblum: Ah great, didn't know about that.  I updated in the following commit to use uuid4: https://github.com/ibpsa/project1-boptest/pull/444/commits/bcbbf29b049a8332a3907e9de6f06b11d26c07aa.

dhblum:
@kbenne I think this is ready to merge once tests pass. Added unit test argument to /submit API which saves a payload json that is checked for unit tests instead of sending to dashboard.  Unit tests also check invalid arguments and whether the time period scenario has completed or not.

Let me know any other comments if you have them.

Issue #406, opened 2022-02-23
Add support for Spawn


kbenne:

dhblum:
@kbenne Is the commit https://github.com/ibpsa/project1-boptest/commit/fd538ccee25e6a6b755f9030ba3fa4e9e382b77f all that's needed for this issue?  Can I cherry pick it to the main branch to close this?  Or, any update in what's in the commit now (e.g. spawn version?)

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/447.

PR #447, opened 2022-07-19
issue 406 spawn simulation support

dhblum:
This is to address #406 to add simulation support for test case FMUs using Spawn, by installing Spawn into the BOPTEST run-time environment docker container.  

This does not address tools/workflows for compiling test case FMUs with Spawn.


dhblum:
Ready to merge once tests pass.

Issue #450, opened 2022-07-22
Reduce travis unit test job time


dhblum:
The ``make test_all`` unit tests run within a single job are approaching the 2 hour time limit for travis as seen here: https://app.travis-ci.com/github/ibpsa/project1-boptest/jobs/577415219.  This issue is to either reduce the run time of the ``make test_all`` job, or to split the build into multiple jobs, for instance running each test in its own job.

dhblum:
@kbenne FYI opening this issue to address.

kbenne:
Good deal. It seems like there should be plenty of low hanging fruit that will yield much more efficient testing. I wonder if the strategies diverge between service and master? I'd rather not diverge if possible.

dhblum:
Just took a quick look at https://github.com/ibpsa/project1-boptest/blob/issue448-service-submit/testing/makefile.  If we split the tests into individual jobs, I think what I proposed in https://github.com/ibpsa/project1-boptest/blob/split_travis_jobs/.travis.yml should work for service branches too.  This doesn't make the tests more efficient, just breaks jobs up into smaller chunks so they don't run into timeout problems.  But the whole test build will take longer to complete.  There's some optimization there available, like I'm not sure both the jm and boptest_base image need to be built for every test necessarily.

The alternative is to make the the one job with all tests more efficient.  If for Service, one of these things means provisioning all at once in the beginning, I don't think it's much of a divergence than already exists.  Already, the Service ``makefile`` uses a provision command (e.g.  ``cd .. && make provision TESTCASE=multizone_residential_hydronic``) within individual tests which doesn't exist in main.  These could be removed and combined into a single command within ``test_all`` at the beginning, similar to use of ``build_jm_image`` and ``build_boptest_image`` there.  Similar can be done in main for things like building the javascript and julia images.  But, in both cases, then the individual tests can't be run independently, which could be annoying for unit testing during development (so don't always have to run all tests).  Or, we put in sort of if-then clauses to check what case is being run.

kbenne:
In the interest of wrapping up 0.3.0 I'm favoring a two prong approach. For the near term, I think I'd like to bring your split approach over to service.

After we wrap 0.3.0 I'd like to pursue building out one provision at the start of the whole thing, with some kind of logic to keep development convenient.

dhblum:
[...]

PR #452, opened 2022-07-22
issue 450 reduce test time

dhblum:
This is to address #450.  It reduces time for ``test_all`` by 
- only building julia and javascript controller images once instead of twice each for testcase1 and testcase 2.
- simulating scenario tests in a single step instead of the default step, which can be one minute or one hour depending on the case.


dhblum:
Refer to #454.

PR #453, opened 2022-07-26
issue 450 travis time 2

dhblum:
This is for #450 and just implements the compilation of javascript and julia images once, as opposed to #452 which tries a few additional things.


dhblum:
Refer to #454.

Issue #465, opened 2022-08-05
testcase: multizone_office_simple_hydronic


icupeiro:
this issue is to track the two-zone office model based on real practice experience by DeltaQ

icupeiro:
@dhblum the model is currently working with MSL4.0.0. 

How should we tackle this issue? Is it possible to combine it with the work in #422 or should I revert to MSL3.2.3 for the moment?

dhblum:
What compiler are you using with MSL4?

icupeiro:
I used Modelon Impact to create and simulate the model so far

dhblum:
I aim to tackle #422 this coming year to switch the default compiler we use for unit testing from JModelica to OpenModelica, as much as possible at least following on improvements to OpenModelica being made this year to simulate Buildings Library models.  Do you know if the model simulates with the nightly build of OpenModelica?  For now I'll also add your model to the tracking in #422.

dhblum:
And how close is the model to being done?  Are there signal exchange blocks, test case .csv data, and documentation all done?  I ask because I'm trying to gauge how quickly you're trying to or would be ready to add this model to the repo, compared to how long #422 will take to complete.

icupeiro:
hey Dave,

I just tried to compile it in OM, unsucessfully but so far due to the more strict requirements of OM with the Modelica standards. I will have a look at this

Test case .csv data is there, doc and exchange blocks are missing so far but I just had a meeting with @JavierArroyoBastida to discuss these points, so they be there soon.

The idea is to have something fully ready by the next IBPSA meeting

icupeiro:
Update: signal exchange blocks are ready, model compiles with OM (but simulation fails at initialization)

icupeiro:
Follow-up update: after adding some energy dynamics (changing from steady-state) in some components here and there, the model now compiles and SIMULATES in OM (despite the results I checked preliminary differ)

icupeiro:
Update on compilation process:

Tried to substitute line 45 in parser.py from compile_fmu() to give the path of the FMU compiled in OM
Failed due to an error loading the FMU binaries

`
Traceback (most recent call last):
  File "compile_fmu.py", line 39, in <module>
    fmupath = compile_fmu()
  File "compile_fmu.py", line 33, in compile_fmu
    fmupath = parser.export_fmu(modelpath, [mopath])
  File "/usr/local/testing/parsing/parser.py", line 215, in export_fmu
[...]

PR #469, opened 2022-08-08
Issue465 multizone office simple hydronic

icupeiro:
This tackles #465 

Pending to-do: 

- [x] Discuss model in next IBPSA meeting (fan/pump powers, control, humidity, etc...)
- [x] Write documentation

@dhblum @JavierArroyoBastida FYI. You should be able now to toy around with the model and run the tests

I will ask Lieve to allocate time for the next IBPSA meeting to present the model.

Note that in the makefile I did not include the compilation step in the tests due to the incompatibility with MSL4.0.0. This step is done manually using OCT at the moment


JavierArroyoBastida:
Great to see this. Thanks @icupeiro!

icupeiro:
Seems that I forgot a few commits from master into my branch, which changed the makefile to a more generic script
Note that, consequently, the tests will fail since the model cannot compile. 
I had to run them in my own machine after deploying the testcase.

icupeiro:
@JavierArroyoBastida ready for review

Note: the compilation of the model will fail since MSL4 is being used. The parser script should be adapted (or compilation should be kept out of the makefile)
I ran the tests manually without the makefile.

testcases/multizone_office_simple_hydronic/models/library_versions.json
```diff
@@ -0,0 +1,4 @@
+{
+"IDEAS_COMMIT": "cd7ce8c44d597c6ca02a0db4c90cbf2bed29754d",
+"BUILDINGS_COMMIT": "ea147aeee6e0a0239f1b48f5cd75f7e72963bf73"
+}
```
dhblum, line 4: @icupeiro Are these commits corresponding to specific library version tags (e.g. Buildings v9.0.0) or are they master branch commits between versions?

dhblum:
@icupeiro Thanks a lot for all this.  Looks quite complete.  I will review as well, and need to think a minute on how best to handle the testing and compiler issues.

testcases/multizone_office_simple_hydronic/models/library_versions.json
icupeiro: These are the latest master branches (from yesterday). In fact, IDEAS was recently updated so that some material properties were modified using better sources. See https://github.com/open-ideas/IDEAS/commit/cd7ce8c44d597c6ca02a0db4c90cbf2bed29754d. 

Needless to say, documentation has been properly updated.

icupeiro (commented):
@dhblum have a look here. I had to adapt https://github.com/ibpsa/project1-boptest/blob/master/data/get_html_IO.py to be compliant with the new restapi

icupeiro:
> @icupeiro Thanks a lot for all this. Looks quite complete. I will review as well, and need to think a minute on how best to handle the testing and compiler issues.

My personal feeling is to follow the single responsibility principle of clean code (each function/test should do just one thing and do it well), therefore splitting compilation (if really needs to be tested, which can be another discussion) and individual tests

icupeiro (commented):
this change adds percentiles readblocks as suggested by @saroele 

In pratice, a single zone/space will have many sensors. By using the mean temperature of them as the tracking setpoint, it will happen that the RBC/MPC will try to keep this mean temperature close the lower/upper boundary. However, by following this boundary with the mean, many of the space sensors will be in discomfort.  

Therefore, what we do in practice is to introduce temperature percentiles in the set of sensors within a space to minimize the real discomfort. The lower percentile will be used for heating and upper percentile for cooling. 

We rudimentary mimicked this percentiles based on our data experience by adding a +/- 1 K temperature band to the space temperature.

These blocks are not used for computing KPIs (but please have a thought about whether they should). Should we add them to the documentation somehow?

JavierArroyoBastida:
Thanks a lot @icupeiro, this emulator is excellent work and a great addition to the BOPTEST framework. I've added my review in [this PR](https://github.com/icupeiro/project1-boptest/pull/1). Notice the file in `/doc/reviews/BOPTEST Test Case Peer Review Document_MulZonOffSimHyd_rev1.docx`. There are a few minor edits that I've implemented myself. What I suggest in the review document are only minor edits as well.

icupeiro:
@JavierArroyoBastida your comments have been included with the exception of point 3, i.e. renderization of some symbols in doc/index.html. The html code seems good and I can see it using Linux Mint Vanessa 21 + Firefox 105.0 and also at the documentation pane using OpenModelica build 1.20.0~dev-305-geb5e7e0. So I assume the rendering issue comes from the browser client you are using?

JavierArroyoBastida:
Thanks @icupeiro! yes, I've double-checked and it works in chrome indeed, I was using PyCharm before. So I have no further comments. Maybe @dhblum can have a look already.

icupeiro:
@dhblum just a gentle reminder to have a look/merge this issue!

dhblum:
Thanks a lot @icupeiro.  I'm aiming to review in the coming week.

dhblum:
@icupeiro I'm reviewing.  Ok if I push some changes to your fork?  So far, unit test related.

icupeiro:
Sure! No problem

icupeiro:
@dhblum reminder. It would be nice if this gets merged before I depart

dhblum:
@icupeiro At our last IBPSA BOPTEST progress meeting I discussed with @saroele a concern I have about merging without an open-source compiler option available to continue maintenance and testing afterwards.  We discussed doing additional testing with OpenModelica to see if it can compile an FMU that can be simulated in BOPTEST, and if not what the issues are.

icupeiro:
@dhblum the model compiles with OM, I just had troubles finding the target platform option in OM and I moved forward with OCT. But compilation of the FMU should be fine.

dhblum:
@icupeiro That's good it complies in OM.  I would like the whole workflow to be tested though, including simulation of the resulting FMU in BOPTEST to identify any remaining issues.

icupeiro:
To-do:

- [ ] Modify heat producer to convert it to an ASHP
- [ ] Update to latest IDEAS 
- [ ] Compile the model with OCT
- [ ] Run BOPTEST tests

dhblum:
Closing and replacing by https://github.com/ibpsa/project1-boptest/pull/618.

Issue #463, opened 2022-08-03
Abstract test case input output lists in unit tests


dhblum:
This is to abstract the definition of input and output lists within if-else statements in the API unit tests in ``utilities.py`` to the ``setUp`` method in each test cases's specific testing file ``test_<testcase>``.  For example in these locations:

- https://github.com/ibpsa/project1-boptest/blob/master/testing/utilities.py#L565
- https://github.com/ibpsa/project1-boptest/blob/master/testing/utilities.py#L689
- https://github.com/ibpsa/project1-boptest/blob/master/testing/utilities.py#L716
- https://github.com/ibpsa/project1-boptest/blob/master/testing/utilities.py#L865

dhblum:
Assign @EttoreZ.

EttoreZ:
I will make a pull request on the branch Issue463_abstract_testcase_inout_lists_unittests

EttoreZ:
I made the pull request, I only have one comment:
In the script  `utilities.py` function `test_advance_false_overwrite`, the input used for `bestest_hydronic` is different than in function `test_invalid_advance_value`, namely `u = {'oveTSetSup_activate': 0, 'oveTSetSup_u': 273.15+60,              'ovePum_activate': 0,'ovePum_u': 1}` and `u = {'oveTSetSup_activate': 0, 'oveTSetSup_u': 273.15+60,'ovePum_activate': 0}`. In the pull request I used the first input dictionary and running the tests passed all of them, since the actual values or number of inputs do not matter for those specific unit test.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/472.

PR #471, opened 2022-08-09
Issue463 abstract test case tests

EttoreZ:
This pull request is to address issue #463.


testing/test_bestest_air.py
```diff
@@ -60,6 +60,8 @@ def setUp(self):
         self.url = 'http://127.0.0.1:5000'
         self.step_ref = 3600
         self.test_time_period = 'peak_heat_day'
+        self.input = {'fcu_oveTSup_activate': 0, 'fcu_oveTSup_u': 290}
```
dhblum, line 63: @EttoreZ I suggest calling "input_false_overwrite" or add a comment that the value of activate is meant to be 0 for the test_advance_false_overwrite API test.

dhblum (changes requested):
@EttoreZ Thanks for this PR.  Please address the one inline comment and then I think this will be ready to go.

EttoreZ:
Thank you @dhblum for the careful review. I added the comment `#<u_variable>_activate is meant to be 0 for the test_advance_false_overwrite API test` above the self.input definition and updated the pull request.

dhblum (approved):
Thanks @EttoreZ for the udpates.

Issue #475, opened 2022-08-18
Error message if fmu path not existant


EttoreZ:
This issue is to check that `fmupath` in `testcase.py TestCase __Init__` method exists or else an exception is thrown.  The proposed modification is to add a `try: execpt:` with the `os.path.exists()` method.

dhblum:
Sounds great.  Want to make a PR to https://github.com/ibpsa/project1-boptest/tree/issue475_fmuPath?

EttoreZ:
Thank you for creating the branch @dhblum . I will make a PR on that branch tomorrow morning

EttoreZ:
@dhblum I updated the pull request. Also to leave trace here for our conversation today. In a future version the error trigger should be moved directly to the bash script to be more precise in the error identification.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/478.

PR #476, opened 2022-08-19
Issue475 fmu path

EttoreZ:
This pull request is for issue #475. A slight modification was added with the respect to the original implementation. `os.path.exists()` will return True even is the testcase name is wrong and does not exists. This happens because the docker-compose command will create an empty directory instead. For this reason an additional check was added as `os.path.isfile()` to check that `wrapped.fmu` is an actual existing file


dhblum:
Thank you @EttoreZ!

PR #478, opened 2022-08-24
Issue475 fmu path

dhblum:
This is for #475.


dhblum:
Will merge when tests pass.

Issue #398, opened 2022-01-28
Provide a list of points to /results


dhblum:
This issue is to suggest making the ability to provide a list of points to the /results API call, instead of requiring one at a time.  The return would be the same object, just with additional arrays of point data.  

While I acknowledge the potential memory issues of calling for large amounts of data (e.g. https://github.com/ibpsa/project1-boptest/issues/240), this would reduce the need for excessive network requests (say in a long for-loop), which could otherwise bottleneck the system.  And, the user would have control over the amount of data returned at one time if memory becomes too much an issue.

dhblum:
We may also consider similarly for forecast data when addressing https://github.com/ibpsa/project1-boptest/issues/356.

EttoreZ:
@dhblum would you mind create the branch issue398_ListOfPointsResultsAPI so I can make a pull request? All the unit tests passed on bestest_air and bestest_hydronic. However, there are still a couple of things to check before the pull request is completely ready:

- [ ] Check that python examples run correctly
- [ ] Check that Juilia examples run correctly
- [ ] Make change for Javascript examples
- [ ] Check comments and documentation

dhblum:
Thanks @EttoreZ.  Branch created here https://github.com/ibpsa/project1-boptest/tree/issue398_ListOfPointsResultsAPI.  Make the PR and unit tests will check bullets 1 and 2.  3 shouldn't be too hard to add if I have to, and I can review comments/docs.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/480.

PR #479, opened 2022-08-24
Issue398 list of points results api

EttoreZ:
This pull request is for issue #398. However, as mentioned in the issue it still needs a couple checks and changing the Javascript example before it can be merged.


testing/utilities.py
```diff
@@ -371,7 +371,7 @@ def results_to_df(self, points, start_time, final_time, url='http://127.0.0.1:50
 
         df = pd.DataFrame()
         for point in points:
-            res = requests.put('{0}/results'.format(url), data={'point_name':point,'start_time':start_time, 'final_time':final_time}).json()['payload']
+            res = requests.put('{0}/results'.format(url), data={'point_names':[point],'start_time':start_time, 'final_time':final_time}).json()['payload']
             df = pd.concat((df,pd.DataFrame(data=res[point], index=res['time'],columns=[point])), axis=1)
```
dhblum, line 375: I think the for-loop and concatenation can be removed since now the points are provided in a list all at once.

EttoreZ: I updated the pull request removing the for cycle and also found and corrected a small bug.

PR #480, opened 2022-08-25
Issue398 list of points results api

dhblum:
This is for #398 and makes the argument of ``/results`` a list of points.


dhblum:
I think ready to merge if tests pass.  @JavierArroyoBastida Mind taking a look?  @kbenne this would require an update on boptest-service API.

testcase.py
```diff
@@ -662,29 +662,27 @@ def get_results(self, point_name, start_time, final_time):
...
-
+            for point_name in point_names:
+                # Get correct points
+                if point_name in self.y_store.keys():
+                    payload[point_name] = self.y_store[point_name]
+                elif point_name in self.u_store.keys():
+                    payload[point_name] = self.u_store[point_name]
+                else:
+                    status = 400
+                    message = "Invalid value {} for parameter point_names.  Check lists of inputs and measurements.".format(point_names)
```
JavierArroyoBastida, line 675: Maybe printing `point_name` (instead of `point_names`) is more useful here. This would tell the user directly which variable does not match the list of inputs and outputs.

JavierArroyoBastida (approved):
Looking good to me. I just added a small suggestion. Thanks @EttoreZ and @dhblum for the implementation!

testcase.py
dhblum: Good point, thanks.  Done in https://github.com/ibpsa/project1-boptest/pull/480/commits/b79c9c408666e3d0db80187fd12f6978c9ad8f4d.

dhblum:
Thanks @JavierArroyoBastida for the review!

PR #491, opened 2022-09-21
Issue409 two zone hydronic apartment

EttoreZ:
I updated the pull request following @dhblum comments.


dhblum:
Thank you @EttoreZ!

Issue #496, opened 2022-11-02
Update two zone hydrone model documentation


EttoreZ:
As per title this issue is to update the documentation of the two zone hydronic model referencing the newly published [paper](https://www.tandfonline.com/doi/full/10.1080/19401493.2022.2120631). @dhblum can you create a branch so I can make a pull request?

dhblum:
Created here: https://github.com/ibpsa/project1-boptest/tree/issue496_twoZoneDocUpdate.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/498.

PR #497, opened 2022-11-03
Issue496 two zone doc update

EttoreZ:
This pull request is for updating the documentation of the two zone hydronic apartment model according to #496.


dhblum:
Thanks @EttoreZ for the PR.  One comment though, I would prefer adding the full bibliographic citation to the end of the documentation section (e.g. end of the "Architecture" section) so it can be read and known without a reader or viewer having to click on the link.

EttoreZ:
Thanks @dhblum for the comment. I updated the documentation with the whole citation.

dhblum (approved):

Issue #529, opened 2023-03-31
Unit tests depends on list order that is not determintistic


kbenne:
This line and the very next one depends on a determinisitic order of object keys

https://github.com/ibpsa/project1-boptest/blob/b040c4283476ba3e7db76d70d43c789946ff3627/testing/utilities.py#L604

While testing BOPTEST-Service I discoverd this test failing due to the nondeterminism. (Service appears to be sending the keys back in the oposite order) I think the test should not assume a particular order.

https://stackoverflow.com/questions/42232200/as3-deterministic-json-string-from-an-object

@dhblum

dhblum:
Thanks for pointing out @kbenne.  I agree, and see it's because of ``forecast_points[0]`` in the next line 605.  I suggest updating to specify the one point to fetch in the test setup for each test case as is already done for singular input and measurement checks, for example as in https://github.com/ibpsa/project1-boptest/blob/b040c4283476ba3e7db76d70d43c789946ff3627/testing/test_bestest_air.py#L54.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/530.

PR #530, opened 2023-03-31
Issue529 che for lis ord

dhblum:
This is for #529.  It specifies a specific forecast point to check for each test case in associated unit test.


dhblum:
This should be ready to merge if tests and review pass.

dhblum:
Merging after discussion with @kbenne.

Issue #540, opened 2023-05-11
SyntaxError: JSON.parse: unexpected character at line 1 column 268 of the JSON data for testcase2


kefeimo:
### Description (How to reproduce)

For
* version 0.4.0-dev
* TESTCASE=testcase2

When using `curl http://127.0.0.1:5000/kpi`,
![image](https://github.com/ibpsa/project1-boptest/assets/28743873/e7a695d6-3e9f-4f56-adad-99d06e127b05)

it returns raw json
```json
{"message":"Queried KPIs successfully.","payload":{"cost_tot":0.005221328521720328,"emis_tot":0.013053321304300819,"ener_tot":0.026106642608601637,"idis_tot":0.0,"pdih_tot":null,"pele_tot":0.037526602248043876,"pgas_tot":null,"tdis_tot":2.0119814359592807,"time_rat":NaN},"status":200}
```
Note the return is not a valid json due to `"time_rat":NaN`

### Expected Result
Return valid Json.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/543.

PR #542, opened 2023-05-17
fix issue #540

Enderdead:
The issue #540 reports an invalid JSON format when a KPI indicator is requested at the beginning of a scenario. This occurs because the kpi_calculator.get_computational_time_ratio function returns an np.nan value when no time ratio is available. Since the Flask JSON encoder cannot convert np.nan to null, this results in an incorrect JSON format.

I propose to return an arbitrary zero value from get_computational_time_ratio if no data is available for computation. This fixes issue #540 with a very minor change.

### DIFF
--- a/kpis/kpi_calculator.py
+++ b/kpis/kpi_calculator.py
@@ -694,7 +694,7 @@ class KPI_Calculator(object):
         '''
 
         elapsed_control_time_ratio = self.case._get_elapsed_control_time_ratio()
-        time_rat = np.mean(elapsed_control_time_ratio)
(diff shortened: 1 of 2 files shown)


dhblum:
Thank you very much @Enderdead for the PR!  I think a 0 is an ok solution, though a bit misleading.  What about returning a ``null``?

dhblum:
I'd prefer a Python ``None`` object.  If you can update this PR for that, that'd be great.  Can you also add a brief line to the ``releasenotes.md`` under the backward-compatible changes?  I'm going to change the target merge branch for this PR to a staging branch and will update unit test reference results after you make your updates.

Enderdead:
Done.

PR #543, opened 2023-05-26
Issue540 nan kpi

dhblum:
This is for #540.  In particular, it fixes the KPI calculator so that it returns a Python None object for the computational time ratio KPI instead of NaN if the test has not been advanced yet.

Verified using Mozilla Firefox browser for testcase2 after starting test case but not advancing at all, as exampled in #540 to demonstrate error:
Screen Shot 2023-05-26 at 6 19 05 PM

Unit tests should pass.

@Enderdead can you verify I updated ``contributors.md`` properly for you?  And thank you for the fix.


Enderdead:
Good enough. Thank you for the reference.

Issue #533, opened 2023-04-11
Incorrect check for none in testcase.py


dhblum:
In ``testcase.py`` in ``/advance`` in Line https://github.com/ibpsa/project1-boptest/blob/master/testcase.py#L279, the second check (``and u[key]``) was originally implemented to check for an input value of ``None``.  This is because Flask sends the value of inputs that are not specified by the user as ``None``.  This also works if the value is of type string, which the Flask API set up in ``api.py`` provides.  However, if the value is of type float, as in the case of the BOPTEST-Service API and when using the ``json`` attribute of python requests from a client, and = 0, then this will cause the check to fail and the value will not be included in the overwriting.  

Related to https://github.com/ibpsa/project1-boptest/issues/528 and https://github.com/ibpsa/project1-boptest/issues/532.

FYI @kbenne

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/544.

PR #544, opened 2023-05-31
Issue533 che key

dhblum:
This is for #533.  In particular, it:
- Fixes the handling of an overwrite input to the testcase.TestCase.advance of value 0 and delivered not as a string (as in the case of using BOPTEST-Service and the ``json`` attribute of python requests package, instead of the ``data`` attribute).
- Implements a unit test that replicated the error and contains reference results for the fix
- Adds a new unit test module for testing ``testcase.py`` directly
- Makes a couple docstring edits


dhblum:
@kbenne This would be good to test on a test Service deployment.

Ye-Chuang:

Issue #550, opened 2023-06-24
Python example testcase1_scenario returns a value error in Python3


kuzha:
When running the Python-based example controller testcase1_scenario.py using Python3 (tested in Python3.6 and 3.9), it will return the following error: "ValueError: Out of range float values are not JSON compliant". However, it works well in Python2.

The issue occurs from the line 145 in examples/python/interface.py, where np.inf is assigned to the final_time for the /results API request done later.  The np.inf is probably not json serializable.

In the meantime, we can use any float value for that assignment as a workaround, for example, assigning a value equal to 14 days later than the start time since all scenarios are two-week periods, so that final_time = start_time + 14*86400.

dhblum:
@kuzha When I test with python 3.9.5 I actually don't get this error, but I do with 3.9.0  The unit test ``.travis.yml`` specifies python "3.9", which I think might pull a 3.9.x version where x > 0, which is why our unit test for this doesn't catch the error.  I suggest we actually update the python version of unit tests to 3.10 or 3.11 at this point.  I will create a branch that does this and see if unit tests still pass.

dhblum:
@kuzha Can you confirm you don't see the error if tested with python >=3.9.5?

dhblum:
Update unit test python version to 3.11.4 here: https://github.com/ibpsa/project1-boptest/tree/issue550_scenarioExample.

dhblum:
Also suggest updating readme as in here: https://github.com/ibpsa/project1-boptest/commit/1d8f3968a9bdb3cddbc174fcf6700b764e167b80.

kuzha:
Hi @dhblum,
I tested this example in Python3.6.9, Python3.9.5 and Python3.9.16 and all did not work with the same error message. 
I then checked the packages in travis, which specifies pandas==1.2.5 numpy==1.20.2 matplotlib==3.3.4 requests==2.25.1. 
I noticed that my package versions are all different from those packages. After a few trials, I identified that it is the package "requests" gives errors. When I install "requests" using pip install, the versions I got are all higher than 2.25.1. I then installed the version 2.25.1 for requests, the example works well without errors for all the three python versions I have.
You could also have a check at the "requests" version you have in your pythons. If they are all 2.25.1, then try to update to a higher version and see if you get the same error as I had.

dhblum:
[...]

Issue #545, opened 2023-06-14
GET/results Enabling default start_time and final_time 


kefeimo:
## Tag
Feature Required/ User Experience Improvement

## Description
As mentioned in https://github.com/ibpsa/project1-boptest/blob/fef02e961dc8ebaaa199ad1580a9cf50c8723e51/docs/workshops/BS21Workshop_20210831/Introduction_to_the_BOPTEST_framework.ipynb?short_path=0c1a91e#L748-L749, to achieve simulation results, we can use the following code snippet.

```
args = {'point_names':['reaTZon_y'],'start_time':-np.inf, 'final_time':np.inf}
res = requests.put('{0}/results'.format(url), data=args).json()['payload']
```

Note that we need to use `-np.inf` and `np.inf` to specify the whole range of the timestamp. Since np (aka numpy) is not native library of python, it introduces extra dependency. 
In addition, arguably, it is easier for a user to get a full timestamp of the simulation result and filter it afterward rather than specifying it beforehand. Plus, it is not clear what would be the available start_time and final_time when using PUT/scenario.

## Expected/proposal behavior
When leaving start_time, final_time empty, the RestAPI returns the full timestamp of the simulation result (i.e., set start_time==-np.inf, final_time==np.inf as default)

PR #553, opened 2023-07-04
Issue550 scenario example

kuzha:
#550 
@dhblum 
I checked the travis [documentation](https://docs.travis-ci.com/user/languages/python/) and my understanding is that the current stable python in travis is version 3.9. So I decided to stick to version 3.9. 
In Python 3.9, I upgraded pip first and then upgraded the four packages _numpy_, _pandas_, _matplotlib_ and _requrests_. Please double check if you get the same version as I modifed in travis.
Other changes are:
- [x] final_time and steps in interface.py
- [x] readme
- [x] release notes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,50 +23,50 @@ before_script:
 
 jobs:
   include:
-  - python: 3.11.4
-    install: pip install --upgrade pip && pip install pandas==1.2.5 numpy==1.20.2 matplotlib==3.3.4 requests==2.25.1
+  - python: 3.9
+    install: pip install --upgrade pip && pip install pandas==2.0.3 numpy==1.25.0 matplotlib==3.7.1 requests==2.31.0
     script: cd testing && make build_jm_image && make test_parser
-  - python: 3.11.4
-    install: pip install --upgrade pip && pip install pandas==1.2.5 numpy==1.20.2 matplotlib==3.3.4 requests==2.25.1
+  - python: 3.9
+    install: pip install --upgrade pip && pip install pandas==2.0.3 numpy==1.25.0 matplotlib==3.7.1 requests==2.31.0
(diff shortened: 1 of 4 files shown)


dhblum:
Thanks @kuzha.  I'll double check the versions.  But looks like in the unit tests we're getting an error: ``ImportError: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with 'OpenSSL 1.0.2g  1 Mar 2016'. See: https://github.com/urllib3/urllib3/issues/2168``.  I have a feeling this is related to the Ubuntu version travis is running, which looks to be Ubuntu 16 (Xenial): https://github.com/ibpsa/project1-boptest/runs/14753430755.  I think it's worth updating to 18 at least if not 20 via https://docs.travis-ci.com/user/reference/linux/.

kuzha:
@dhblum I googled a bit about this error and had the same conclusion as you: the error comes with the openSSL version in Ubuntu 16. I changed the default Ubuntu version in .travis.yml to 18 just to give it a try to see if the test can pass.

I have only installed Ubuntu 18 on my VM so I don't know if everything works on 20. I can give it a test locally first if you would like to update to 20.

kuzha:
Some tests passed but some failed in CI.
The same error occurs due to the new _requests_ package, where more cases using _np.inf_ have been identified. Most notably in utilities.py (see below), which has been used in many tests.

>  File "/home/travis/build/ibpsa/project1-boptest/testing/utilities.py", line 639, in test_set_get_scenario
>     df = self.results_to_df(points_check, -np.inf, np.inf, self.url)

Not sure what is the best way forward. @dhblum

dhblum:
Thanks @kuzha.  Hm ok.  This is starting to look related to https://github.com/ibpsa/project1-boptest/issues/545.

PR #558, opened 2023-07-13
Issue554 forecast horizons of 0

### LINKED ISSUE #554: Forecast horizon = 0
In [this line](https://github.com/ibpsa/project1-boptest/blob/a9aaec3bb44bc9bd362f3d3002e6d940584c708e/testcase.py#L862) we prevent from forecast horizons of 0. Fundamentally, that makes sense. However, the only way for a user to get to know current boundary condition data other than weather is by using the forecaster with a prediction horizon of 0. In the case of weather data, we have the weather station block, but there are others like pricing or setpoints that couldn't be retrieved for the current time otherwise. A use case can be e.g. a PI controller that reacts to deviations from a setpoint, or a reinforcement learning agent that includes current setpoints and pricing in its observation space. I therefore suggest to change the `<=` of that line to `<`.

JavierArroyoBastida:
Fixes #554.

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -6,6 +6,7 @@ Released on xx/xx/xxxx.
 
 **The following changes are backwards-compatible and do not significantly change benchmark results:**
 
+- Allow forecast horizons of 0 to retrieve boundary condition data at current time like pricing or temperature setpoints. This is for [#554](https://github.com/ibpsa/project1-boptest/issues/554).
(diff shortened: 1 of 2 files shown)


dhblum:
Thanks @JavierArroyoBastida.  This looks good.  I just had one suggested change which I made directly to your fork to see if that commit also re-triggered the CI tests.  It looks like the Travis CI - Branch is still "expected" as it was before too.  I'll have to look into why this is the case if it doesn't resolve itself during that time.

dhblum (approved):

JavierArroyoBastida:
I see. Thanks for looking into it. Hopefully, tests are triggered now. I pushed both commits rather fast, which might have been the source of the problem yielding the branch as "expected".

PR #559, opened 2023-07-19
Issue554 forecast horizon0

dhblum:
This is for #554 following PR #558 to see if it triggers the Branch CI.

FYI @JavierArroyoBastida

### DIFF
--- a/releasenotes.md
+++ b/releasenotes.md
@@ -6,6 +6,7 @@ Released on xx/xx/xxxx.
 
 **The following changes are backwards-compatible and do not significantly change benchmark results:**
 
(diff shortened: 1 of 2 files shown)


dhblum:
@JavierArroyoBastida Hm, ok, looks like both travis builds are triggered now.  So there are two required checks for ibpsa:master that CI be run on the branch and also that it be run in the case the pull request is merged.  But, it seems the branch test is only trigged in travis if the branch is already within ibpsa, seemingly as opposed to a fork as was done in PR #558.  I'm learning something new and this behavior is good to know.  You are included in the team to write directly to the ibpsa repo, so in the future you may just create a branch and PR from ibpsa instead of your fork?

dhblum:
Yes, usually folks sending from forks are not as experienced or part of the core development team, so I've actually created staging branches on ibpsa to merge against from a fork and then finalize any cleanup on that branch instead of pushing directly to someone's fork.

Issue #547, opened 2023-06-15
BACnet Interface


dhblum:
This issue is to make a generalized BACnet interface available for BOPTEST.  It will be based on what's been started in https://github.com/epaulson/boptest-bacnet-proxy, with at least the following extensions:

- Usable for all test cases
- Not reliant on a semantic model (expansion of semantic support in BOPTEST planned for later work)
- Unit testing

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/567.

PR #567, opened 2023-08-25
Issue547 bacnet

dhblum:
This is for #547.  It:
- Adds /bacnet with an interface to deploy a bacnet device with bacnet objects for each input and measurement boptest point and translate reading/writing requests made through bacnet to that device to/from the running boptest test case.  Also adds readme to that directory.
- Adds a simple example script for reading/writing through bacnet and uses the script in a new unit test, which deploys a boptest test case, deploys the bacnet interface, and confirms a write and read request successfully are conducted.
- Updates license for added code basis being under MIT
- Adds a .ttl file to each test case /models directory which specifies how to instantiate bacnet objects for upon deployment of the bacnet device.  Also adds a script to /bacnet which can be used to automatically create the .ttl for any test case that can be run in boptest.
- Updates test case design doc

To Do:
- [x] Review
- [x] Update release notes

@epaulson mind taking a review?


epaulson:
Sorry that I took so long to get to this - I think it looks good! It nearly ran out of the box on my mac (there's no hostname -I on macOS so test_bacnet.py needed a 1 line change, though I just hardcoded in my IP to get past it) 

I think for now just calling everything a brick:Point is fine until there's smarter semantic data that can come out of BOPTEST and actually set the point types correctly, as well as exporting data about the equipment and spaces to make the Brick model richer. 

I will probably archive my repo and just point folks here.

dhblum:
@epaulson Thank you for reviewing!  Glad it looks good.   To your comments:

- Do you know of the alternative for  hostname -I on macOS?  Then, we could detect the OS and write the correct command, so that someone could run on a mac.  It's not totally essential as test_bacnet is meant to be a CI test which is run on Ubuntu in travis.  
- Yes, that was my thinking in the use of brick:Point.  Fuller featured semantic modeling is to be worked on in the coming year.

epaulson:
There's an 'ipconfig' command that works but you need to know which interface you're asking about (which is usually en0 on the mac, but not always?) The more portable option is to open a socket and then ask what IP address got attached to that socket, which will sort through whatever defaults are at play (and so should pick up the same defaults that docker using) - but as you mentioned, since that's only a CI test that will usually be on linux it's probably not a big deal.

dhblum:
Ok thanks @epaulson, I suggest keeping how it is then to not overcomplicate the matter.

dhblum:
This is ready to merge when tests pass.

PR #572, opened 2023-09-04
Gh pages custom

JavierArroyoBastida:
This is to automate the workflow of generating the website content using GitHub Actions.


dhblum (approved):
@JavierArroyoBastida This is really great, thank you again.  I've tested and looks like things work well.  I don't see any issues.  I will merge and set up the repo to generate the website using the github actions and see if everything goes well.

Issue #576, opened 2023-09-13
Reduce unit testing time by removing some test case time period scenario tests


dhblum:
The inclusion of larger and more complex test cases increases the time it takes to run simulations.  In addition, for larger test cases, use of CS fmus is proving to take longer to run simulations.  For example, the latest CI tests in https://github.com/ibpsa/project1-boptest/pull/536 take ~7 hours to complete and it's possible the CI tests in https://github.com/ibpsa/project1-boptest/pull/569 would take a day or more given reported computation times on that emulator.  

A significant amount of CI time is running all iterations of all time period scenarios of each test case for their three-week duration (warmup week + two-week test period).  For emulators with heating and cooling systems, that equates to 5 tests (peak/typical_heating/cooling and mix).  For emulators without one or the other, there are two-day tests that are run to test on results for specific seasons not covered by a time period scenario (e.g. summer, winter).

My proposal here is to reduce the amount of simulations run for each test case as they may be overkill and unnecessarily adding to testing time.  This could be done perhaps by either:

1. Reducing which time period scenarios that are tested.  For example, only run peak_heating and peak_cooling, or similar, instead of all.
2. Only run the two-day seasonal tests for all emulators, which would significantly shorten things.

The down sides here are less test coverage on the test case models and not having the KPIs and results for every time period scenario for every test case tested with each CI run.  Note that I'd still intend KPIs and results would still be tested for any time period scenarios or seasonal tests that are run.

I wanted to float this and see if there are thoughts for or against or alternative ideas.

PR #584, opened 2023-11-03
Issue 576 Reduce Unit Test Time

dhblum:
This reduces the unit test time by shortening the simulation of scenarios for each test case to two days instead of the whole two weeks.

This reduces testing time by about 50 min and the size of the reference results directory by about 120 MB.


dhblum:
This is for #576.

PR #588, opened 2023-11-16
Gh pages custom

JavierArroyoBastida:
@dhblum please let me know if something like this is what you had in mind, or if you have further suggestions/edits. You should be able to preview in [my fork](https://javierarroyobastida.github.io/project1-boptest/) once it builds.


dhblum:
Thanks a lot @JavierArroyoBastida!  Yes this is something like what I had in mind.  My current suggestions are:

- Order from most recent at the top
- Switch the wording a bit for titles so that they start with a more distinguishing feature/event so it's easier to sort:  e.g. "BS2021 Workshop: Introduction to the BOPTEST Framework."  Also, then when we have future workshops we can advertise and post registration information as well.
- Perhaps divide into two sections more explicitly: "User Tutorials and Workshops", "Software Development Guidance" (which will have only one item right now).

dhblum:
Awesome, thank you!  I'm going to merge, and then just make a couple other website updates I noticed regarding the Service update.

Issue #239, opened 2020-10-04
BOPTEST crashes when forecast outside simulation year is requested


haraldwalnum:
When performing simulations up to 1 year, the model crashes when the requested forecast exceeds the one year data in the resource files.  I guess this is due to pandas failing when asked for data outside the index.
I see some possible solutions to this:
1.	Continue by reading data from the top of the dataframe when the bottom is reached. However, this might results in big discontinuities for some of the weather data.
2.	Keep the last value constant. This should be a stable solution, but not realistic. At least for longer simulation periods, the impact should be small
3.	Repeating data for the last day of the year. There could still be some discontinuities but less than for solution 1.

EttoreZ:
@haraldwalnum  I also found the same issue. In theory the IBPSA library does what suggested in step 1., if typical weather data is used the discontinuity shouldn't be too big, but it could still meddle with the controller. I guess an idea could be use point 1. I also report the error log that I obtain:

  _File "/home/developer/.local/lib/python2.7/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/developer/.local/lib/python2.7/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/developer/.local/lib/python2.7/site-packages/flask_restful/__init__.py", line 468, in wrapper
    resp = resource(*args, **kwargs)
  File "/home/developer/.local/lib/python2.7/site-packages/flask/views.py", line 89, in view
    return self.dispatch_request(*args, **kwargs)
  File "/home/developer/.local/lib/python2.7/site-packages/flask_restful/__init__.py", line 583, in dispatch_request
    resp = meth(*args, **kwargs)
  File "restapi.py", line 142, in get
    forecast = case.get_forecast()
  File "/home/developer/testcase.py", line 427, in get_forecast
    interval=self.interval)
  File "/home/developer/forecast/forecaster.py", line 75, in get_forecast
    plot=plot)
  File "/home/developer/data/data_manager.py", line 317, in get_data
    data_slice_reindexed.loc[:,key] = f(index)
  File "/usr/lib/python2.7/dist-packages/scipy/interpolate/polyint.py", line 79, in __call__
    y = self._evaluate(x)
  File "/usr/lib/python2.7/dist-packages/scipy/interpolate/interpolate.py", line 610, in _evaluate
[...]

PR #599, opened 2023-12-13
Issue239 extend forecast

EttoreZ:
This pull request is to address issue #239, related to simulation across the year.


dhblum:
Thanks @EttoreZ.  I just merged a pending PR that causes conflict now in the release notes.  Can you resolve the conflict?

EttoreZ:
> Thanks @EttoreZ. I just merged a pending PR that causes conflict now in the release notes for this PR. Can you resolve the conflict?

Sure, thank you for the heads up @dhblum.

dhblum (commented):
@EttoreZ Thank you for this.  The implementation splits the data at year-end to 1) before the end of the year inclusive of the last data point at the end of the year (midnight), and 2) after end of the year not inclusive of the first data point at the start of the year (midnight).  Please confirm this is what we agreed should be the case, rather than what is suggested in [this comment](https://github.com/ibpsa/project1-boptest/issues/239#issuecomment-1474436978).  This might be the case but I can't find it in my notes.  The implementation might make sense if this is what we agreed so that the forecast is more consistent for any interval through the full first year if a user only intends to simulate one year.  Can you please confirm and remind me?  Note I've plotted the results of unit tests below to show the performance of the implementation for forecast intervals of 1800 and 123 (as in the unit tests), compared to the reference boundary condition csv file (for an example relative humidity).

forecast_over_year_check

Ye-Chuang:

EttoreZ:
> @EttoreZ Thank you for this. The implementation splits the data at year-end to 1) before the end of the year inclusive of the last data point at the end of the year (midnight), and 2) after end of the year not inclusive of the first data point at the start of the year (midnight). Please confirm this is what we agreed should be the case, rather than what is suggested in [this comment](https://github.com/ibpsa/project1-boptest/issues/239#issuecomment-1474436978). This might be the case but I can't find it in my notes. The implementation might make sense if this is what we agreed so that the forecast is more consistent for any interval through the full first year if a user only intends to simulate one year. Can you please confirm and remind me? Note I've plotted the results of unit tests below to show the performance of the implementation for forecast intervals of 1800 and 123 (as in the unit tests), compared to the reference boundary condition csv file (for an example relative humidity).
> 
> forecast_over_year_check

@dhblum thank you for the careful review. The implementation was indeed changed with the respect to the one mentioned in the comment for that exact reason, so that a 1 year simulation would be consistent.

dhblum:
@EttoreZ Ok I see thank you!

dhblum (changes requested):
@EttoreZ This is looking good, thanks.  Can you please add a description of the behavior, potentially with figure similar to what I posted in the discussion, to the design documentation here? https://github.com/ibpsa/project1-boptest/blob/master/docs/design/source/forecasts.rst.  In particular, you may add a subsection "Getting Forecasts Across Year-End"

dhblum (approved):
This is ready to merge once tests pass.

PR #602, opened 2023-12-14
Issue239 extend forecast

EttoreZ:


dhblum:
@EttoreZ In the future, to update an existing PR, I think you can just push new commits directly to a development branch (in this case issue239_extendForecast) instead of first making a PR from a development branch on your fork?  Otherwise, we have lots of PRs opening and closing and it is confusing.

Issue #626, opened 2024-02-22
Store scenario results when completed


haraldwalnum:
This issue is to implement a function that stores the scenario results KPIs++ and time series measurements in the simulation directory, so that this can be restored later with the "testi_id" if ran with service.

HWalnum:
@dhblum and @kbenne , we can continue the discussion here.
We should decide what we should store, and naming.
Actually I don't see the need to store results from different scenarios with different names, but rather just overwrite existing stored results.

The need for downloading results from the database is a special need (so far the only use case is competitions?). 
If users need to store results from multiple runs, they should rather use the API to request results and kpis, and store themselves.

Also when results from multiple runs from the same testcase is needed, why not require that the user reloads the testcase and get a new "test_id". I don't think it is much overhead with this?

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/632.

Issue #514, opened 2022-12-20
Track points utilized by test controllers


dhblum:
This issue is to implement a mechanism in BOPTEST to keep track of the inputs, measurement, and forecast points utilized by a test controller.  

A question is if the list of variables for each inputs, measurements, forecasts should be reported back from /kpis, or just the length of the lists, or reported in some other way?

JavierArroyoBastida:
I suggest having the list of accessed inputs, measurements and forecast variables returned by /kpis. I think it is important to return the list of variables, and not just the length of the lists since there are variables that are less likely to have in practice than others e.g. it is clear that accessing a heat flow measurement is less likely than a temperature measurement. That notion is something that we want to make explicit when evaluating the controller.

dhblum:
I agree @JavierArroyoBastida with returning the full list of points utilized as it would be more informative.

SenHuang19:
I am a little concerned that changing this API makes BOPTEST more towards a control development environment rather than a control testing environment as it begins to create differences.

dhblum:
@SenHuang19 Can you elaborate on what you mean by "create differences?"  The intent is to keep track of which sensors, control points, and forecast variables are used by test controllers to help evaluate their relative need for data and actuation authority.

SenHuang19:
Sorry, I didn't make myself clear.
I understand the necessity of having this feature.
This feature doesn't make too many differences and what I commented on is whether we should allow customized settings via API in general.
Strictly speaking, the testing period and even sampling interval should be predefined, and control developers need to include necessary changes inside their control sequence to accommodate those predefined parameters. 
This way, BOPTEST provides a solid foundation for control benchmarking.

dhblum:
The testing period can be chosen using the ``/scenario`` API.  The sampling interval of historic data is specified within boptest to 30 seconds.  Generally, I think I agree with you.  But I'm not clear on what settings you refer to when mentioning "customized settings via API", if not in this issue.

dhblum:
Initial implementation which tracks inputs, measurements, and forecast points used in a controller test and reports in /kpis is here: https://github.com/ibpsa/project1-boptest/tree/issue514_variableTrack.  An example call to /kpi using the bestest_air case as an example with various API requests already made looks like:
```
{
    "message": "Queried KPIs successfully.",
    "payload": {
        "cost_tot": 6.735242989734677e-05,
        "data_forecasts": [
            "TWetBul",
            "TDryBul",
            "relHum"
[...]

PR #629, opened 2024-02-27
Issue626 store results

HWalnum:
@dhblum. Here is a first version of implementation of automatic storing results.
Please comment, both on the implementation and which parameters should be stored in results.json

### DIFF
--- a/testcase.py
+++ b/testcase.py
@@ -1358,9 +1358,12 @@ class TestCase(object):
         # store results_json
         with open(name + ".json", "w") as outfile:
             json.dump(results_json, outfile)
-
+ 
+        # get list of results, need to use output metadata because duplicate inputs are removed
+        result_list = self.input_names + list(self.outputs_metadata.keys())
         # get results trajectories
-        results = self.get_results(self.input_names + self.output_names, self.initial_time, self.end_time)[2]
+        results = self.get_results(result_list, self.initial_time, self.end_time)[2]
+        # convert to dataframe
         results_df = pd.DataFrame.from_dict(results)
         # store
         results_df.to_csv(name + ".csv")
--- a/testcase.py
+++ b/testcase.py
@@ -1145,27 +1145,16 @@ class TestCase(object):
         dash_server = os.environ['BOPTEST_DASHBOARD_SERVER']
         # Create payload
         uid = str(uuid.uuid4())
-        payload = {
-          "results": [
-            {
-              "uid": uid,
-              "dateRun": str(datetime.now(tz=pytz.UTC)),
-              "boptestVersion": self.version,
-              "isShared": True,
-              "controlStep": str(self.get_step()[2]),
-              "account": {
+        test_results = self._get_test_results()
+        api_parameters = {
+            "uid": uid,
+            "isShared": True,
+            "account": {
                 "apiKey": api_key
-              },
-              "forecastParameters":{},
-              "tags": tags,
-              "kpis": self.get_kpis()[2],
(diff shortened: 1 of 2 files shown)


testcase.py
```diff
@@ -1327,6 +1330,44 @@ def _get_full_current_state(self):
...
+
+
+        Returns
+        -------
+        None.
+
+        '''
+
+        name = "results"
+        # get results_json
```
dhblum, line 1346: To reduce code duplication, I suggest abstracting the creation of a results.json structure in a separate function for all fields that are common with what is submitted to dashboard, defined here https://github.com/ibpsa/project1-boptest/blob/master/testcase.py#L1143.  Then, ``post_results_to_dashboard`` and ``store_results`` can share use of that function to build an initial json and add additional fields as-needed  (if any) for submission to dashboard or writing to file, respectively.

testcase.py
```diff
@@ -1327,6 +1330,44 @@ def _get_full_current_state(self):
...
+        be retrieveable with the test_id
+
+
+        Returns
+        -------
+        None.
+
+        '''
+
+        name = "results"
```
dhblum, line 1345: Suggest calling "file_name" instead of "name"

dhblum (changes requested):
Thank you @HWalnum.  Please see my inline comments and address.  Also, can you add a note to the ``releasenotes.md`` under the backwards compatible section?  And can you re-target this PR to ``ibpsa:issue626_storeResults`` instead of ``ibpsa:master``?

testcase.py
```diff
@@ -1327,6 +1330,44 @@ def _get_full_current_state(self):
...
+        # get results_json
+        results_json = {
+            "dateRun": str(datetime.now(tz=pytz.UTC)),
+            "boptestVersion": self.version,
+            "emulatorName": self.get_name()[2]['name'],
+            "controlStep": str(self.get_step()[2]),
+            "forecastParameters": {},  # for future use to store used parameters?
+            "measurementParameters": {},  # for future use to store used parameters?
```
dhblum, line 1353: I have an initial implementation described here: https://github.com/ibpsa/project1-boptest/issues/514, but ran into an issue for truly accurate tracking as described at the end of that issue.

dhblum:
Noting this is for https://github.com/ibpsa/project1-boptest/issues/626.

testcase.py
HWalnum: I proposed a solution. However, I had some trouble with submitting results to the dashboard, with both previous and updated solution, so I have not checked that it works properly.

testcase.py
HWalnum: Ok. I support the last proposed solution. But I guess this could be implemented first anyway?

testcase.py
dhblum: @HWalnum I'll take a look, but looks like you got a couple results submitted for BESTEST Hydronic Heat Pump under the account "hwaln" at https://dashboard.dev.boptest.net/.

HWalnum: Thanks @dhblum. I think those are from some extra tests I did with the service.

dhblum:
This is looking good.  I will merge this to the staging branch and make a few adjustments there.

testcase.py
dhblum: Ok actually I'm having some trouble submitting to the dashboard too.

dhblum: And Ok I see.  I'm meeting with Kyle tomorrow morning for a few things and he can help work out what's going on.

Issue #500, opened 2022-11-09
Add check and warning in data_manager.py for all weather variables


EttoreZ:
This issue is to add a warning message in [`data/data_manager.py` ](https://github.com/ibpsa/project1-boptest/blob/master/data/data_manager.py), that checks wether all the common weather variables, taken from [`data/categories.json`](https://github.com/ibpsa/project1-boptest/blob/master/data/categories.json), are included in the `testcase/Resources/weather.csv`. In the case `weather.csv` does not include all variables, a warning will be displayed to the user.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/645.

PR #645, opened 2024-05-14
Issue500 update data manager

EttoreZ:
This PR addresses #500, a warning was added to the `data/data_manager.py` script if any of the weather variables present in `data/categories.json` is missing from `wrapped.fum/resources/weather.csv`.

### DIFF
--- a/data/data_manager.py
+++ b/data/data_manager.py
@@ -427,7 +427,8 @@ class Data_Manager(object):
 
         # Get zone and boundary data keys allowed
         zon_keys, bou_keys = self._get_zone_and_boundary_keys()
-
+        weather_keys = list(self.categories['weather'].keys())
+        
         # Initialize test case data frame
         self.case.data = \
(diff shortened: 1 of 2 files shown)


dhblum (changes requested):
@EttoreZ I think this functionality should go in the method ``save_data_and_jsons`` since that is invoked when the csvs are moved to the FMU when compiling the test case and is when other similar warnings are shown to a test case developer if there are issues with the different files.  

``load_data_and_jsons`` is called when the test case docker container is deployed and the warning message will be lost among alot of other container deployment messages, not known to a typical boptest user (who is not a test case developer), and may not be displayed at all if the container is deployed in the background or a remote server (i.e. on aws).

EttoreZ:
@dhblum thank you for the comments, they all make sense. 

- I moved the warning to the method `_append_csv_data` within `save_data_and_jsons `, which is where the .csv files are checked.
- I updated the release notes according to the suggestion.

PR #495, opened 2022-10-06
Issue494 baselinetesting

terrancelu92:
The testcase driver scripts for running baseline controllers and associated results files are included.

Issue #494, opened 2022-10-04
BOPTEST baseline controller testing


terrancelu92:
This issue is for the BOPTEST baseline controller testing and results.

dhblum:
Development branch created at https://github.com/ibpsa/project1-boptest/tree/issue494_baselinetesting.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/651.

PR #651, opened 2024-06-11
Issue494 baselinetesting 2

dhblum:
This is for #494 and followup up on https://github.com/ibpsa/project1-boptest/pull/495 to finalize merging.

To-Dos:
- [x] Verify added if-else in interface.py (see [here](https://github.com/ibpsa/project1-boptest/pull/651#discussion_r1635583251))
- [x] Pass unit tests
- [x] Add ``/baselines`` directory description to structure in ``README.md``
- [x] Final functionality testing


examples/python/interface.py
```diff
@@ -136,14 +136,26 @@ def control_test(control_module='', start_time=0, warmup_period=0, length=24*360
...
         res = check_response(requests.put('{0}/scenario'.format(url), json=scenario))['time_period']
         print(res)
-        # Record test simulation start time
-        start_time = int(res['time'])
-        # Set final time and total time steps to be very large since scenario defines length
-        final_time = np.inf
-        total_time_steps = int((365 * 24 * 3600)/step)
+        
+        # Initialize test with a scenario without time period specified
+        if res == None:
```
dhblum, line 144: @terrancelu92 Can you remind me, why did you add this if-else, since you can specify the ``scenario`` parameter as ``None`` and go to line 160, which initializes the simulation according to start_time and warm_up period instead of with a scenario.  This if-else between lines 144-150 seems to duplicate the functionality.

terrancelu92: @dhblum, I vaguely recall there being a reason for not setting the `scenario `parameter to `None` and going to line 160. Let me look into it and get back to you. Thanks for bringing this to my attention.

dhblum: @terrancelu92 Ah ok, I see the issue, and think your approach is ok.  I tried thinking a bit if there was a more streamlined approach, but couldn't come up with anything particularly better.  I think I will just adjust the comments around the code though to make it more clear to a reader what the if-else's are for.

Issue #617, opened 2024-02-06
Migrate to BOPTEST-Service as only deployment architecture


dhblum:
This issue will migrate [BOPTEST-Service](https://github.com/NREL/boptest-service) to be used as the only deployment architecture and RESTful API.  The primary advantage and reason for this is to alleviate the need to maintain the two closely matched but not exact RESTful APIs, alleviate developers and users having to develop and maintain interfaces for each of them, and generally streamline the implementation of new BOPTEST features which are also supported by BOPTEST-Service.  A downside is that the service architecture is more complex for those who use BOPTEST locally.  However, with continued usage of docker-compose, users will still be able to deploy BOPTEST easily and in the same basic way.  

Todos:
- [x] Migrate BOPTEST-Service code (Service v0.4.0, BOPTEST v0.6.0)
- [x] Fix 400 error response message in Service when wrong point name - see https://github.com/ibpsa/project1-boptest/issues/671.
- [x] Fix Julia example controller variable error, silence docker build
- [x] Pass all BOPTEST unit tests
- [x] Remove javascript example
- [x] Migrate Service-specific unit tests
- [x] Update release notes
- [x] Update README, including README example commands and unit test (if still needed)
- [x] Incorporate lessons-learned from ADRENALIN (logging and timeouts) - see https://github.com/ibpsa/project1-boptest/issues/684.
- [x] Update get_html_IO, find_days, create_ttl to work with Service architecture (see also https://github.com/ibpsa/project1-boptest/issues/700)

dhblum:
@kbenne FYI I've created this issue.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/652.

Issue #634, opened 2024-03-13
Drop support for example Python test controllers for Python 2


dhblum:
Currently, unit tests run ``make test_python2`` in a Python 2 environment (https://github.com/ibpsa/project1-boptest/blob/master/.travis.yml#L93) to check the example Python controllers run in Python 2.  This issue is to drop that support by removing the tests from the testing makefile and travis .yml.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/642.

PR #607, opened 2024-01-21
Pin Dockerfile image to `linux/x86_64`

mattrobmattrob:
Pin the base Docker image to `linux/x86_64` to enable the image to be used on Apple Silicon machines easily. This can be reverted when PyFMI is distributed to `linux/arm64` (https://github.com/modelon-community/PyFMI/issues/212).

Error for future reference:
```
$ docker-compose up
...
1.589 2024-01-22 16:51:21 (54.1 MB/s) - '/home/user/miniconda.sh' saved [74403966/74403966]
1.589
1.600 PREFIX=/home/user/miniconda
1.816 Unpacking payload ...
1.822 rosetta error: failed to open elf at /lib64/ld-linux-x86-64.so.2
1.822  /home/user/miniconda.sh: line 352:    38 Exit 141                extract_range $boundary1 $boundary2
1.822         39 Trace/breakpoint trap   | "$CONDA_EXEC" constructor --extract-tarball --prefix "$PREFIX"
------
failed to solve: process "/bin/sh -c cd $HOME && \twget https://repo.anaconda.com/miniconda/Miniconda3-py310_23.1.0-1-Linux-x86_64.sh -O $HOME/miniconda.sh && \t/bin/bash $HOME/miniconda.sh -b -p $HOME/miniconda && \t. miniconda/bin/activate && \tconda update -n base -c defaults conda && \tconda create --name pyfmi3 python=3.10 -y && \tconda activate pyfmi3 && \tconda install -c conda-forge pyfmi=2.11 -y && \tpip install flask-restful==0.3.9 werkzeug==2.2.3 && \tconda install pandas==1.5.3 flask_cors==3.0.10 matplotlib==3.7.1 requests==2.28.1" did not complete successfully: exit code: 133
```

The conda binary is already pinned to x86_64 so this just goes one step further with the base image:

https://github.com/ibpsa/project1-boptest/blob/358358566fc167e06c3957f5f79a6cef46812365/Dockerfile#L22

Fixes https://github.com/ibpsa/project1-boptest/issues/608.

PR #652, opened 2024-07-01
Issue617 refactor service with history 2

dhblum:
This is for #617 and #671 and in place of #622.  It does everything from #622, but merges the Service architecture in a way to include the [nrel/boptest-service](https://github.com/NREL/boptest-service) git history.


Ye-Chuang:

dhblum:
@kbenne FYI.

dhblum:
@EttoreZ @javiarrobas @terrancelu92 This PR is ready for beta testing.  Can you checkout the branch ``issue617_refactorService_with_history2`` (https://github.com/ibpsa/project1-boptest/tree/issue617_refactorService_with_history2) and test deployment and use of BOPTEST using the migrated service architecture?  You should be able to start with the README and go from there :) Let me know if you're able to do some testing, and report back any issues you find, preferably within the next 1.5 weeks.

FYI @kbenne

service/.gitignore
javiarrobas, line 1: Wouldn't be best to have this in the `.gitignore` at the root directory?

service/README.md
javiarrobas, line 1: I suggest merging this one with the `README.md` at the root directory.

service/version.txt
javiarrobas, line 1: I think we don't need to track the service version anymore after it is merged.

testing/makefile
```diff
@@ -272,12 +287,32 @@ test_testcase:
...
-	cd .. && TESTCASE=testcase2 docker-compose up -d
+	cd .. && docker compose build -q web worker provision	
+	cd .. && docker compose up -d --quiet-pull web worker provision
+	python sleep10.py
+	curl http://127.0.0.1:80/measurements
+	curl http://127.0.0.1:80/forecast
+	curl http://127.0.0.1:80/advance -d '{"oveTSetRooHea_u":293.15,"oveTSetRooHea_activate":1, "oveTSetRooCoo_activate":1,"oveTSetRooCoo_u":298.15}' -H "Content-Type: application/json"
+	cd .. && docker compose down
+
+test_python2:
```
javiarrobas, line 298: I don't think we're still testing Python 2. We can remove this already.

javiarrobas (approved):
Nice work, it's looking good to me. I just went through it and tried to understand as much as possible. While reading I added a couple of minor comments/suggestions.

javiarrobas:
I've been trying to deploy it myself locally. It seems that upon calling `docker compose up web worker provision` it provisions only for `bestest_hydronic`, `multizone_office_simple_hydronic`, and `test_multizone_residential_hydronic` testcases. Any idea why?

image

Would it be possible to specify just one testcase instead of provisioning multiple? When working locally, we typically want to interact with just one test case. If we select just one test case to deploy we could save some time and resources.

testing/makefile
javiarrobas: See: https://github.com/ibpsa/project1-boptest/issues/634.

javiarrobas:
Another suggestion is to pin all Dockerfile images to `linux/x86_64` as we did [here](https://github.com/ibpsa/project1-boptest/pull/607) to enable cross-platform builds. This is particularly relevant for those running on Apple silicon.

service/.gitignore
kbenne: I think `service/.gitignore` can be removed. I think most (all?) of this file is irrelevent now and anything not covered by the existing toplevel `.gitignore` could be added as neeeded.

service/README.md
kbenne: No argument from me.

kbenne:
> I've been trying to deploy it myself locally. It seems that upon calling `docker compose up web worker provision` it provisions only for `bestest_hydronic`, `multizone_office_simple_hydronic`, and `test_multizone_residential_hydronic` testcases. Any idea why?
> 
> image
> 
> Would it be possible to specify just one testcase instead of provisioning multiple? When working locally, we typically want to interact with just one test case. If we select just one test case to deploy we could save some time and resources.

I'm not sure why some of the testcases are not provisioned. From your log it does appear to be true, but I would double check by calling the `GET /testcases` endpoint, which will tell you for sure which test cases were provisioned.

The [default provision command](https://github.com/ibpsa/project1-boptest/blob/891f99baafc471fc1520e2490d5085e481cf4d9f/service/provision/Dockerfile#L12) which is invoked by `docker compose up provision` is `CMD [ "python3", "-m", "boptest_submit", "--shared", "--path", "./testcases/"]`. Note the `--path` argument. The [script will glob](https://github.com/ibpsa/project1-boptest/blob/891f99baafc471fc1520e2490d5085e481cf4d9f/service/provision/boptest_submit/boptest_submit.py#L20)  for FMUs at or below the given path. It is possible to provision a single test case, although it isn't convenient and maybe we should probably do something to make it better. To provision a single test case you would need to bring up the service with `docker compose up web worker` and then from a separate terminal run `docker compose run --no-deps provision python3 -m boptest_submit --shared --path ./testcases/multizone_office_simple_hydronic/`. Note that you can reprovision a test case, so for instance if you are in a development loop working on a test case, you could make your changes, recompile the FMU, and call the provision command again, which will replace the existing test case without needing to bring the service down and back up.

javiarrobas:
> I'm not sure why some of the testcases are not provisioned. From your log it does appear to be true, but I would double check by calling the `GET /testcases` endpoint, which will tell you for sure which test cases were provisioned.

I'm double checking with the `GET /testcases` endpoint and I can confirm that other test cases are not provisioned:
image

Notice that it gets stuck with `test_multizone_residential_hydronic` (not with `multizon_residential_hydronic`). Not sure if that has to do with anything, but I'm mentioning just in case: 

image

My guess is that it does get stuck in the `for` loop when provisioning the `test_multizone_residential_hydronic` test case, probably because of me running in a Mac M1 laptop with the worker configured with `linux/x86_64` and thus running under emulation. I've encountered issues with that in the past. If that's the case, I'm not sure there is much we can do about it. It seems there are now pyfmi binaries for `linux-aarch64` see [here](https://github.com/modelon-community/PyFMI/issues/212), so unpinning the worker image might work, but honestly I don't think it will because the FMUs are still shipping `linux/x86_64` binaries.

service/README.md
dhblum: Ok, but what do you think about in the API request table, somehow having some separation in the requests specific to user/namespace-specific test cases from those to select from the public IBPSA test cases and other functional API requests (e.g. /advance, /results, etc.)?  Because I think of those as more advanced user type things that I don't want people to get confused about when just starting to try BOPTEST with our public test cases.

I could propose something.

service/version.txt
dhblum: @kbenne Ok, please make a commit to take care of this.

service/.gitignore
dhblum: Removed in https://github.com/ibpsa/project1-boptest/pull/652/commits/aedb62293e671e12fe152bb07429a96e3bd2189e.

testing/makefile
dhblum: Agreed, but hm, this should have been removed in https://github.com/ibpsa/project1-boptest/pull/642 and subsequent master merge.  But removed now in https://github.com/ibpsa/project1-boptest/pull/652/commits/aedb62293e671e12fe152bb07429a96e3bd2189e.

service/.gitignore
dhblum: I mean removed in https://github.com/ibpsa/project1-boptest/pull/652/commits/7530d591a33a1ddec93d131d6ef6178739d6c72c.

dhblum:
> > I'm not sure why some of the testcases are not provisioned. From your log it does appear to be true, but I would double check by calling the `GET /testcases` endpoint, which will tell you for sure which test cases were provisioned.
> 
> I'm double checking with the `GET /testcases` endpoint and I can confirm that other test cases are not provisioned: image
> 
> Notice that it gets stuck with `test_multizone_residential_hydronic` (not with `multizon_residential_hydronic`). Not sure if that has to do with anything, but I'm mentioning just in case:
> 
> image
> 
> My guess is that it does get stuck in the `for` loop when provisioning the `test_multizone_residential_hydronic` test case, probably because of me running in a Mac M1 laptop with the worker configured with `linux/x86_64` and thus running under emulation. I've encountered issues with that in the past. If that's the case, I'm not sure there is much we can do about it. It seems there are now pyfmi binaries for `linux-aarch64` see [here](https://github.com/modelon-community/PyFMI/issues/212), so unpinning the worker image might work, but honestly I don't think it will because the FMUs are still shipping `linux/x86_64` binaries.

@javiarrobas Hm, will need to look into this.  Provisioning actually doesn't start running (or initializing) any of the test cases.  All it's doing is copying the FMU files into the Service's storage for when a request to start a test case is received, the FMU is loaded into the worker, upon which then the initialization happens per previous method of local deployment (e.g. with testcase.py).

dhblum:
@EttoreZ Can you try deploying BOPTEST Service in this branch on your Apple Silicon Laptop?

EttoreZ:
@dhblum I tried deploying on my mac and it worked just fine as you can see below: 
image
I had the same issue as @javiarrobas when I started testing service using the wrong branch `issue617_refactorService` instead of the correct branch `issue617_refactorService_with_history2`.
I was also able to run a test script that initializes, asks for measurements, advances, asks for forecasts, results and KPIs for `bestest_air`, see screenshot below:
image
It is a bit slower than on my Linux machine, probably due to the platform emulation.

service/version.txt
kbenne: [done](https://github.com/ibpsa/project1-boptest/commit/b648373ae72a81789c5e8aa25a59ad579a95ae5d)

javiarrobas:
I've checked again and still get the same output. I previously removed all my boptest-related images to ensure they are rebuilt with the last version. I also confirm I'm using the latest version of `issue617_refactorService_with_history2 ` branch as in commit `b648373a`. @EttoreZ what's your Mac and MacOS versions? what's the command you're running? 
Mine is a MacBook Air M1 running with MacOS 15.0.1 (24A348) and the command I'm running is `docker compose up web worker provision`.

EttoreZ:
Hey @javiarrobas. I used the same branch and commit. The computer is a Macbook Pro M1, but I am still running on Sonoma 14.6.1 (23G93). Docker version 20.10.17, build 100c701. The command I run is exactly the same as yours.

dhblum:
I think it has something to do with how the paths are being parsed in the ``submit_all`` function here: https://github.com/ibpsa/project1-boptest/blob/issue617_refactorService_with_history2/service/provision/boptest_submit/boptest_submit.py#L19, upon the command in the Dockerfile here https://github.com/ibpsa/project1-boptest/blob/issue617_refactorService_with_history2/service/provision/Dockerfile#L12.  The issue is that on @javiarrobas machine, the multizone_hydronic_residential test case fmu path is being invoked with a ``test_`` prefix.

dhblum:
@javiarrobas what docker version are you testing with?

EttoreZ:
I upgraded my macOS to 15.0.1 (24A348) and everything still works fine. On different note I noticed that the scripts `data/find_days.py` and `data/get_html_IO.py` were not updated to work with service. In particular:

- `data/find_days.py` has a simulate method that enters the container and runs a simulation there. So it should be updated to work with the worker container.
- `data/get_html_IO.py` uses told current version of the API to get inputs, measurements, and forecasts.

javiarrobas:
> @javiarrobas what docker version are you testing with?

@dhblum I'm working with `Docker version 27.2.0, build 3ab4256` version. 

Sorry but couldn't figure out what's the issue yet :(

kbenne:
Hey guys. I have an M1 Mac. I'll see if I can reproduce this and find a solution.

dhblum:
Thanks Kyle that would be helpful for additional testing.

kbenne:
@javiarrobas since I can't reproduce I'm trying to think of ways to help troubleshoot. One key thing is that the provision container uses a [volume](https://github.com/ibpsa/project1-boptest/blob/b648373ae72a81789c5e8aa25a59ad579a95ae5d/docker-compose.yml#L16) to bring the test case files into the container. I wonder if there is some stale data in there. Here's some possible debug steps.

1. Run the provision container and look around. What files do you see?
```
docker compose run provision bash
ls testcases
```
2. Make sure that all of the old volumes get destroyed.
```
docker compose down -v
```

dhblum:
Thanks @kbenne for the additional testing.  In some ways, good there were no issues, however, I would like to figure out what's going on here before merging.  @javiarrobas if you can, please see if Kyle's suggestions help?  I would emphasize that the use of ``docker compose down`` is the best, most complete, and preferred way of shutting down the BOPTEST service to ensure there aren't hanging elements that could impact subsequent deployments.

javiarrobas:
I finally could figure out what was going on. The suggestion of @kbenne turned out helpful for that. When running the provision container and inspecting the files I got:
image
which made me realize that the `test_multizone_residential_hydronic` was an empty folder located in my testcases' directory probably as a residual of something I did in the past. This was somehow preventing the provisioning of other test cases. Notice that while the folders were empty, it had the `doc` and `models/wrapped.fmu` subfolders inside: 
image
We might want to implement some type of safety mechanism to provision only specific test cases from a list.

kbenne:
I think we can probably come up with a simple enhancement to the globbing mechanism that is used to find FMUs and make this more robust.

kbenne:
Or yeah.. Like you said @javiarrobas, we could just explicitly enumerate them in a list.

dhblum:
Huzzah! 🎉 Thank you @javiarrobas for continuing to try.  One thing I'm curious about is still why it got stuck if there was a valid .fmu in that directory?  Regarding the point of explicit list, we may think about if that list should be in any way accessible so anyone who wants to load in their own test case when running on their machine/setup can.  But that goes back to my initial question, I wonder why it got stuck.

javiarrobas:
> One thing I'm curious about is still why it got stuck if there was a valid .fmu in that directory? 

That's because it was not a valid .fmu file but an empty folder called `wrapped.fmu`. Don't ask me why it was there though because I don't know 😅

dhblum:
Ohh ok, I see!

service/README.md
dhblum: Proposed in recent commits https://github.com/ibpsa/project1-boptest/pull/652/commits/319b215724c89812a4875cac6eba8f4de8a3eeeb, https://github.com/ibpsa/project1-boptest/pull/652/commits/f2901558bbf66e8f98a5c28b7165c6312c321343, and https://github.com/ibpsa/project1-boptest/pull/652/commits/8bc7abd422fc10c42acfdba7f9b17c8e3a8d6eb2.

dhblum:
@javiarrobas  @kbenne @EttoreZ I've addressed all Javier's comments (including a proposed readme convergence), Kyle's implemented updates from ADRENALIN, and Ettore's helped with updating some utility scripts to work with the new architecture.  

I think this PR is very very close to being ready to merge.  Nothing more comes to my mind to be done at this point, except for a protection that was mentioned against what happened with Javier, but perhaps we save that for a specific issue on it's own as an enhancement?

Can you all review/test a final time(s) and let me know any further comments or issues to be addressed?  If you're ok with merging, please approve the github review.  Note that I expect pending tests to pass.

javiarrobas (approved):
Looking good! I agree to address the safeguard mechanism for the issue I had in a separate issue.

EttoreZ (approved):
The pr looks good to me and it’s ready to be merged.

PR #654, opened 2024-07-08
Issue465 multizone office simple hydronic

icupeiro:
@dhblum as requested


icupeiro:
Please revise because it seems there were some old commits from January that were not merged in this branch.
Looking at the files changed, it seems OK to me. There are many format adaptations depending on whether I used OM (simulation) or Dymola (actual coding/modeling work)

Issue #660, opened 2024-07-28
Change position of living room setpoint overwrite block for Multizone Residential Hydronic testcase


EttoreZ:
This issue is to address a problem with the overwrite input to the living room heating setpoint. This setpoint is used to control the boiler for the whole apartment and if overwritten ,`oveTSetPum`, only the pump signal is updated, while the old setpoint is used to control the boiler as highlighted in the figure below: 
<img width="598" alt="image" src="https://github.com/user-attachments/assets/d34b555f-759c-4395-8386-a9e0c13dbed3">
So the connections within the test case model will be updated such that `oveTSetPum` will become `oveTSetPumBoi` and overwrite both the pump and boiler living room setpoints.

dhblum:
CLosed by https://github.com/ibpsa/project1-boptest/pull/667.

Issue #653, opened 2024-07-04
Recompile wrapper.fmu


fsievers93:
Hi there,
I am trying to recompile the MultiZoneResidentialHydronic testcase with an explitcit time stepping scheme.
as far as I understood, I need to generate the FMU from the wrapped.mo file:
project1-boptest//testcases/multizone_residential_hydronic/models/wrapped.mo

At first, I thought I'd need to use the compile_fmu.py. However, this requires the pymodelica libraray, which seems to be quite outdated and is not available on pip or conda. Is there an other way to get it? I also couldn't find a gitRepo. The other option, that I see, is OpenModelica. Here I encountered the issue that I could not find the in MultiZoneresidentialHydronic/package.mo referenced Construciton and ARTEMIS library. Trying to generate the FMU without those libraries results in the following errors:

"""
[326] 14:22:21 Translation Error
[Buildings.Fluid.Sources.BaseClasses.Outside: 83:22-109:86]: Variable weaBus.TDryBul not found in scope Outside.

[327] 14:22:44 Scripting Error
[AppData/Roaming/.openmodelica/libraries/Buildings 8.0.0/Fluid/Sensors/MassFlowRate.mo:20:7-31:97:writable] Error: No matching function found for String(/*Real*/ mod.pumEmiSystem.senMasFlo.m_flow, leftjustified = /*Boolean*/ false, significantDigits = /*Integer*/ 3).
Candidates are:
  String(enumeration(:) $e, Integer minimumLength = 0, Boolean leftJustified = true) => String
  String(Integer $i, Integer minimumLength = 0, Boolean leftJustified = true) => String
  String(Boolean $b, Integer minimumLength = 0, Boolean leftJustified = true) => String
  String(Real $r, Integer significantDigits = 6, Integer minimumLength = 0, Boolean leftJustified = true) => String
  String(Real $r, String format) => String
....
[348] 14:26:49 Translation Error
Initialization problem is structural singular. Please, check the initial conditions.

[352] 14:26:49 Translation Error
Internal error IndexReduction.pantelidesIndexReduction failed! System is structurally singular and cannot be handled because the number of unassigned equations is larger than the number of states. Use -d=bltdump to get more information.

[353] 14:26:49 Translation Error
Internal error Transformation Module PFPlusExt index Reduction Method Pantelides failed!

[354] 14:26:49 Translation Error
[...]

PR #667, opened 2024-08-06
Issue660 update over write block multi res hydro

EttoreZ:
This PR is to address #653 by simplifying pump control and #660 by allowing setpoint change to control the boiler on and off signal.


dhblum (changes requested):
This looks mostly good.  Need to:

- [ ] Check why model takes so much longer to run and whether still reasonable.  For example, unit test ``make test_multizone_residential_hydronic ARG="-s test_typical_heat_day,test_summer"`` takes [45 minutes](https://app.travis-ci.com/github/ibpsa/project1-boptest/jobs/624853443) in this PR while the last master build took [20 minutes](https://app.travis-ci.com/github/ibpsa/project1-boptest/jobs/624750282).  Similar timing discrepancy for ``make test_multizone_residential_hydronic ARG="-s test_peak_heat_day,test_shoulder"`` and ``make test_multizone_residential_hydronic ARG="-s API"``, the latter of which I think exceeds travis wait time of 90 min.
- [ ] Report in release notes what the KPI impacts are for full scenario tests of ``peak_heat_day`` and ``typical_heat_day`` (results from unit tests I've already written are only the first two days).
- [ ] Merge fix from https://github.com/ibpsa/project1-boptest/issues/673 once that issue resolved and update computational time ratio reference results.

dhblum:
@EttoreZ I edited my review after a bit more investigation.

EttoreZ:
> This looks mostly good from a software perspective. But need to:
> 
> * [x]  Check why model takes so much longer to run and whether still reasonable.  For example, unit test `make test_multizone_residential_hydronic ARG="-s test_typical_heat_day,test_summer"` takes [45 minutes](https://app.travis-ci.com/github/ibpsa/project1-boptest/jobs/624853443) in this PR while the last master build took [20 minutes](https://app.travis-ci.com/github/ibpsa/project1-boptest/jobs/624750282).  Similar timing discrepancy for `make test_multizone_residential_hydronic ARG="-s test_peak_heat_day,test_shoulder"` and `make test_multizone_residential_hydronic ARG="-s API"`, the latter of which I think exceeds travis wait time of 90 min.  I observed this when running unit tests locally too.
> * [x]  Once above is investigated, merge fix from [Fix computational time ratio reset #673](https://github.com/ibpsa/project1-boptest/issues/673) (now latest master) and update all reference results.
> * [x]  Report in release notes what the KPI impacts are for full scenario tests of `peak_heat_day` and `typical_heat_day` (results from unit tests I've already committed are only the first two days).

1) I found the issue with the slow down. To avoid having the boiler turn on before the pump does or off after the pump does, causing the simulation to crash, I added a delay of few seconds to the `on` signal for the boiler , and an hold of few seconds on the `off` signal of the pump. These two blocks caused a lot of time events that drastically increased computational time. In the last implementation I removed them, and instead multiplied the control signal of the boiler by a boolean that is 1 if `m_flow > 0.05*m_flow_nom`. Of course if someone starts to control the boiler directly with the overwrite, they might incur in a crash if they turn it on with no flow. After this change the computational time is comparable and 15% lower on an yearly simulation with Jmodelica on the workstation.

2) I merged the latest version of master and rerun the unit tests.

3) I updated the release notes with the percentage difference based on the full `peak_heat_day` and `typical_heat_day` scenarios.

dhblum:
Thanks @EttoreZ.  Can the user turn on the pump without turning on the boiler?  Or will turning on the pump also auto turn on the boiler?

EttoreZ:
If the user turns on the pump it will just enable to boiler, but the boiler will only turn on if the signal from the thermostat/PI control is > 0.  This is because the signals are multiplied with each other.

I see regarding the unit tests, I will be careful.

EttoreZ:
@dhblum This PR is ready for final review and to be merged.

dhblum (approved):

Issue #673, opened 2024-08-21
Fix computational time ratio reset


dhblum:
Upon initialization of a simulation, the array of computational time ratio is reset here: https://github.com/ibpsa/project1-boptest/blob/master/testcase.py#L420.  However, the attribute ``tic_timer`` is not deleted.  Therefore, if an ``/advance`` is made to a test case, and then that test case is initialized using ``/initialize`` or ``/scenario`` and without shutting down the container and deploying a new one, the time to initialize is included in subsequent computational time ratio timer (``tic_timer``) storage for the next test's computational time ratio, since ``tic_timer`` is still present in https://github.com/ibpsa/project1-boptest/blob/master/testcase.py#L264.

This issue will correct this.  The proposed solution is to delete the attribute ``tic_timer`` upon initialization to discard the influence of initialization on computational time ratio calculation for new tests run after such initialization.

This should fix erratic computational time ratio calculation in unit tests seen in https://github.com/ibpsa/project1-boptest/pull/667.  It will influence computational time ratio values calculated in the cases described above.

dhblum:
Just to illustrate the issue:

Starting a container with ``multizone_hydronic_residential`` and logging the time elapsed as ``self.tac_time-self.tic_time`` within the section starting at https://github.com/ibpsa/project1-boptest/blob/master/testcase.py#L264:

Running the unit test for ``peak_heat_day`` (two-days simulation with step of one day) yields only:
Tac-Tic = 0.004599332809448242

Then running the unit test for ``typical_heat_day`` (two-days simulation with step of one day) yields:
First, Tac-Tic: 1513.6133422851562
Then, Tac-Tic: 0.004971504211425781

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/674.

PR #674, opened 2024-08-23
Issue673 time ratio

dhblum:
This is for #673.  It deletes the ``self.tic_time`` attribute of the testcase class upon initialization so as to stop the timer and wait until the first ``/advance`` call to begin again.


dhblum:
@javiarrobas It's a small edit but fixes computational time ratio calculation.  Care to take a look before merging?

javiarrobas (approved):
Good catch. I'm surprised we didn't realize before, but nice to see that it's fixed now!

dhblum:
@javiarrobas Thank you for the quick review!  Yes, glad it'll be fixed.

Issue #684, opened 2024-09-27
Migrate logging and timeout changes from NREL/boptest-service


kbenne:
A handful of changes related to logging and message timeout were made in the NREL/boptest-service repo. These were changes made in support of the Adrenalin compeition. The purpose of this issue is to merge those changes into this repository.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/commit/891f99baafc471fc1520e2490d5085e481cf4d9f.

PR #685, opened 2024-09-27
Issue684 service timeout and logging

kbenne:
This PR addresses #684


docker-compose.yml
```diff
@@ -31,6 +31,7 @@ services:
       - BOPTEST_REDIS_HOST
       - BOPTEST_REGION
       - BOPTEST_TIMEOUT
```
kbenne, line 33: BOPTEST_TIMEOUT is really "idle" timeout. In other words how long there can be no interactions with a running test before it will automatically stop. It might be wise to rename and document these config variables a little better.

service/docs/redis.md
```diff
@@ -3,23 +3,24 @@
...
-| ---------------------------------------- | ----------- | -----------------------------
-| `tests:${testid}`                        | `status`    | `Queued \| Running`
-| `tests:${testid}`                        | `timestamp` | Epoch seconds when test is queued
-| `tests:${testid}`                        | `user`      | The OAuth user "sub", or undefined for anonymous tests
+| Key               | Field       | Value                                                                                    |
+| ----------------- | ----------- | ---------------------------------------------------------------------------------------- |
+| `tests:${testid}` | `status`    | `Queued \| Running`                                                                      |
+| `tests:${testid}` | `timestamp` | Epoch seconds when test is queued.                                                       |
+| `tests:${testid}` | `user`      | The OAuth user "sub", or undefined for anonymous tests.                                  |
+| `tests:${testid}` | `host`      | The value of HOSTNAME for the worker running the test. On K8s this will be the pod name. |
```
kbenne, line 17: There are a number of diffs here, due to formatting adjustment, but the addition of `host` is the actual change. For an instance deployed on Kubernetes, host will be the worker pod name. This is really useful for connecting to a pod that may be running a test and doing diagnostics if there is an issue.

service/test/test_testcase_api.py
```diff
@@ -56,6 +56,9 @@ def test_ibpsa_boptest_testcase():
     response = requests.get(f"{host}/name/{testid}")
     check.equal(response.status_code, 200)
 
+    response = requests.post(f"{host}/advance/{testid}")
```
kbenne, line 59: A lot of API tests in this file, but there was not actaully a test for advance, which is a useful assertion when working on timeout issues.

service/web/server/src/lib/messaging.js
```diff
@@ -39,7 +48,11 @@ class Messaging {
       this.subscribe(responseChannel)
       this.sendWorkerMessage(requestChannel, requestID, method, params)
       responseTimeout = setTimeout(() => {
-        reject(new Error(`Timeout while sending command '${method}' to testid '${workerID}'`))
+        reject(
+          new Error(
+            `Timeout for request: '${requestID}', with method:'${method}', sent to testid: '${workerID}'`,
```
kbenne, line 53: requestID can be (should be) matched to a message received by a worker running a test. Using this log (which will show up in server logs) we can track down, was a message recieved by a worker, and was a response sent. If not, we get this log and can figure out where in the chain it was dropped.

service/worker/jobs/boptest_run_test/job.py
```diff
@@ -91,7 +99,7 @@ def run(self):
     def check_idle_time(self):
         idle_time = datetime.now() - self.last_message_time
         if idle_time.total_seconds() > self.timeout:
-            print("Testid '%s' is terminating due to inactivity." % self.testid)
+            self.logger.info("Testid '%s' is terminating due to inactivity." % self.testid)
```
kbenne, line 102: Using the logger for this stuff will ensure that we get these messages in the log file that clients are able to download.

service/worker/logger.py
```diff
@@ -11,10 +11,12 @@ class Logger:
     """A logger specific for the tasks of the Worker"""
 
     def __init__(self):
-        logging.basicConfig(level=os.environ.get("BOPTEST_LOGLEVEL", "INFO"))
         self.logger = logging.getLogger('worker')
-        self.formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+        self.logger.setLevel(os.environ.get("BOPTEST_LOGLEVEL", "INFO"))
+        fmt = '%(asctime)s UTC\t%(name)-20s%(levelname)s\t%(message)s'
```
kbenne, line 16: This is designed to match the log formatting used in testcase.py. It would be wise to consoldiate the logger implementation at some future point.

docker-compose.yml
```diff
@@ -31,6 +31,7 @@ services:
...
+      - BOPTEST_MESSAGE_TIMEOUT
```
kbenne, line 34: On the other hand. BOPTEST_MESSAGE_TIMEOUT is how long the web server waits for a response from the worker running a test.

PR #688, opened 2024-10-07
Issue 686 & Issue 692 - Examples interface improvements

Jaap-Neven:
Before, this code snippet printed the control step set in the API, but this command is only ran much further down. This resulted in the control step of the previous simulation being printed. Now, the control step given by the user as a function input is directly printed.


dhblum:
Thanks @Jaap-Neven!  One comment: I think it would be safer to print the step that is confirmed to be used within BOPTEST.  I'd recommend after [this line](https://github.com/Jaap-Neven/project1-boptest/blob/issue686_DefaultTimeStep/examples/python/interface.py#L167), adding a line that receives the step (GET /step) and then another line that prints the result.

dhblum:
Thanks @Jaap-Neven.  #692 sounds like a good update, feel free to include the commits in this PR so we can group together all as interface improvements.

dhblum:
Hey @Jaap-Neven thanks again.  Can you add a brief note to the ``releasenotes.md`` summarizing these changes, under the section for "backwards-compatible and do not significantly change benchmark results"?  This is typical for all PRs/changes.  Please use what's there as a template for your addition.

Can you also add your name and affiliation to the ``contributors.md``, which will be a one-time thing for this PR.

Jaap-Neven:
Hi @dhblum, thanks for the feedback, your proposed changes have been made. I'll make sure to change the releasenotes.md in future PRs!

Issue #663, opened 2024-07-31
Update Python Dependency Versions


dhblum:
Numpy 2.0 has been released, there's some nonbackwards compatibility in scipy 1.13 to 1.14 (see https://github.com/ibpsa/project1-boptest/issues/662 and https://github.com/ibpsa/project1-boptest/issues/657), pyfmi 2.13.1 has been released (with support for numpy 2.0).  This issue will update python dependencies, and test out using numpy.trapz in the kpi calculator to eliminate scipy dependence altogether.

Issue #633, opened 2024-03-13
Update unit test testing docker environment from Python 2 to Python 3


dhblum:
Currently, some unit tests (e.g. ``test_parser``, ``test_data``) are run within the ``jm`` docker container environment.  This used to be the same as the boptest test case docker container.  After https://github.com/ibpsa/project1-boptest/issues/146, the test case container uses Python 3, and so the testing environment for these elements should replicate that to avoid needing to use syntax compatible with Python 2 in the framework in order to pass tests.

dhblum:
The approach will be to use the Service ``worker`` container as the unit testing container, so as to test with the same environment as production.  This will be addressed in https://github.com/ibpsa/project1-boptest/pull/690.

dhblum:
Mostly addressed by https://github.com/ibpsa/project1-boptest/pull/690.  However, ``test_data`` still run in ``jm`` due to weather generator.  Suggest splitting out these tests so only test where JModelica is needed uses ``jm``.

PR #690, opened 2024-10-08
Issue663 update python dependancies

EttoreZ:
This PR is for #663 and #633. 

- Removed scipy and matplotlib dependency from worker container.
- scipy.integrate.trapz was substitued with numpy.trapezoid in ``kpis/kpi_calculator.py``, scipy.interp1d linear with numpy.interp, and scipy.inter1d zero with custmo zero hold interpolation in ``data/data_manager.py``.
- Updated worker Python from 3.10 to 3.11, and miniconda version from py310_24.30-1-Linux-x86_64 to py311_24.7.1-0-Linux-x86_64.
- Updated pyfmi from 2.12 to 2.14, update numpy from 1.26.4 to 2.2.1,  update pandas from 1.5.3 to 2.2.3.  
- Update unit tests such that ``test_kpis.py``, ``test_forecast.py``, and ``test_testcase.py`` are run in the ``worker`` docker container, instead of the ``jm`` (JModelca, Python 2) docker container. ``test_data.py`` still runs in the ``jm`` container because it contains testing of ``data_generator.py``, which requires JModelica.


dhblum:
@EttoreZ The error on the first job in travis has to do with the Service-specific tests that were implemented when migrating service over to the ibpsa repo.  Will need to look into that when I review this PR.

dhblum:
This PR will also address https://github.com/ibpsa/project1-boptest/issues/633 as the changes require updating unit testing to Python 3.

dhblum:
@kbenne This PR is looking to update the environment of the ``worker`` container to a new python version, remove a couple package dependencies, and update other packages.    But, in the unit tests, the test on service for the websocket is timing out (see https://app.travis-ci.com/github/ibpsa/project1-boptest/jobs/631555139), therefore failing the tests.  When I test locally, it looks like the websocket tests now take ~16 minutes instead of ~30-45 seconds.    Do you have any idea why the changes in this PR to ``service/worker`` would or could impact the websocket tests like this, to take so much longer?  Or what could be going on here?  I can extend the travis timeout to see if this accommodates, but want to make sure there's not an underlying issue.

kbenne:
The other thing is that I know the websocket is not an official feature.  It kind of came along for the ride with service. If this is too much baggage we could remove it, although up to this point it seemed pretty low effort. Until now maybe.

kbenne:
@dhblum  rereading your comment, I see that it now takes 16 minutes on your local machine!?

kbenne:
yeah, pretty sure I have a fresh clean build. I'll keep poking around though.

dhblum:
Ok.  I tested again with a clean build and got the below using the below command and on Ubuntu 20:

```
docker compose up -d web worker provision && sleep 10 && docker compose up test --no-deps
```
Screen Shot 2025-03-20 at 5 42 57 PM

kbenne:
That is crazy long. These are really basic tests. Pretty much just api checks.

dhblum:
That's why I'm scratching my head.  If I remove the image and rebuild with the master branch and run it, it takes the expected ~1 minute.

kbenne:
how are you removing the image?

kbenne:
Is it possible that the long times are including the build/rebuild step? In my example, I have the images built and ready to go before I `docker compose up...`

dhblum:
I use ``docker rmi`` to remove.  I build first in a separate process, then run the tests.  I can watch the test_testcase_api.py pass in normal time, and then it clearly gets hung up on the socket tests.

kbenne:
I wish I could reproduce this, but as of today I still have not been able to. Notice how the websocket tests are asynchronous. I feel like that is a likely suspect.

One thing to try might be to isolate just one of the tests. You could do that with something like this:
```bash
docker compose run --no-deps test pytest test_websocket_api.py::test_boptest_websocket_one_test
```

In addition, some logging statements after each message with the server might help us figure out where it is getting stuck.

dhblum:
Thanks @kbenne for looking at this again.  I'll try to get back to some more testing this afternoon with your insights.

dhblum:
If I run test_testcase_api.py first and then test_websocket_api.py, it hangs.  But if I switch the order, it works fine.  So there must be something about the end state of the test_testcase_api.py.  I haven't done anymore logging because I tried to implement and then was having pip3 install connection issues to rebuild the docker image.  But I'm going to see if the code below passes the unit tests in travis.

```
docker compose up -d web worker provision && sleep 10 && docker compose run --no-deps test pytest test_websocket_api.py && docker compose run --no-deps test pytest test_testcase_api.py
```

dhblum:
@kbenne ok, specifying the order to run the Service tests such that the websocket tests are run first works ok.  I think I will merge as we have it to not hold this up further.

dhblum (approved):
Based on discussion and additional edits before.

Issue #135, opened 2019-12-07
Add weather forecast uncertainty


dhblum:
This issue is to add weather forecast uncertainty to BOPTEST.  It will start with dry bulb temperature and global horizontal irradiation.

Some todos:
- [x] Fix unit test failure for negative forecast horizon
- [x] Fix unit tests for kpis, forecast, and data
- [x] Add uncertain forecast ref results for test_forecast.py unit test
- [x] Update release notes
- [x] Update contributors
- [x] Update README documentation for scenarios
- [x] Limit interval to hourly, horizon to 48 hours, 
- [x] Update forecast values only hourly for uncertain forecasts
- [x] Update design guide with description
- [x] Add reference in release notes
- [x] Update /forecast readme
- [x] [Add docker --build note to main readme]
- [x] Add options to test case documentation htmls 
- [ ] Add options to test case documentation on website

dhblum:
@laura-zabala Can you make a PR to [this branch](https://github.com/ibpsa/project1-boptest/tree/issue135_forecastUncertainty_2) that adds a couple paragraph description to the design documentation [here](https://github.com/ibpsa/project1-boptest/blob/issue135_forecastUncertainty_2/docs/design/source/forecasts.rst) summarizing the general implementation methods, approach, and related files for the weather forecast uncertainty?  Think of it more as a design specification (focused on implementation) rather than a user guide (focused on API and usage).  I'm thinking you can add a new subsection, something like "Uncertainty Generation" and also add a reference to the now-accepted paper for more details and validation of the methods.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/693.

PR #693, opened 2024-10-18
Issue135 forecast uncertainty 2

dhblum:
This is for #135. A test PR instead of #482 to include changes from Wanfu.


dhblum:
@HWalnum Here's the PR for adding weather forecast uncertainty.  If you can test out, that'd be great.  Note, I'm realizing the README doesn't say, but the values that the temp and radiation uncertainty specs in the /scenario API can take are 'low', 'medium', or 'high'.  We need to add this information, and also to the user guide.

HWalnum:
@dhblum I have done a local test of the forecast uncertainty module. I have not done a review of the implementation. Did you want that as well? You have probably done this already...

In general, it seems to work well. I tested it on a "private" testcase where we have earlier have compared perfect forecast vs an actual weather forecast. Looking at the influence on the performance of the controller (change in mpc objective achievement), the "high" scenario (I tested only with the same scenario for both parameters) gets the most similar results to the "actual weather forecast" case. Looking at the error variance, the "low" scenario for temperature looks more similar to the actual forecast, while the high looks more similar for solar radiation. That said, that is only one case, so I have no objections to the applied distributions. 

Above, you mention a user manual. I could not find that. Is it in the repo?
Looking for this, I found the readme.md in the forecast folder. this should probably be updated.

Lastly, on a separate note. I think it is worth mentioning in the main readme, that when one do changes in the local version of BOPTEST, one needs to add a "--build" flag to the docker compose up statement.

laura-zabala:
Thank you @HWalnum for testing it! 
@dhblum in the case of the temperature, we had more scenarios than for GHI: the four ones from the paper and additionally varios locations around Milan. In that sense, the values of the AR1 model were chosen considering a trade-off between the different sites. I have to check my notes and I'll include here the explanation on how they were estimated. I agree it is good to include this explanation in the guide.

dhblum:
@laura-zabala I've pushed some final edits to the docs.  This PR is just about complete, one unit test is failing which I will fix in the coming days.

@HWalnum I've added code do the following, based on the discussion in the last Project 2 Task 2 meeting.  Care to review/test a final time?
- Restrict users to an ``interval`` of 3600s and ``horizon`` of <= 48 hours when calling for uncertain forecasts in the API
- Update the forecasts, under an uncertainty scenario and when calling for temperature and/or GHI, only at the start of each hour of simulation.

HWalnum:
@dhblum I did some quick tests. It seems to work as intended. 

While it is not unreasonable to require the users to resample the forecast, it would make it simpler for them if we resampled it in the way we do with normal forecasts. Would it not be just to use the interpolation function in the data_manager in the same way, on the error itself, before it is added to the forecast?

I think it is not a requirement, but I think it would make the forecast endpoint more consistent.

dhblum:
Thanks @HWalnum.  I see what you're saying.  So, generate the error based on hourly, and then interpret linearly before adding to the original forecast if the user asks for a different interval.  This would help simplify things for the user, so it's consistent whether asking for temperature and/or solar irradiation or not.  Let me implement something and test it out.

dhblum:
@HWalnum In my latest commit, I made it so an interval not equal to hourly is ok.
For the temperature uncertainty case:

- Error generated first with hourly interval
- Error interpolated to requested interval
- Deterministic forecast generated with requested interval
- Uncertain forecast returned to user by adding error to deterministic forecast
- The result at smaller intervals: a forecast that looks closer to the deterministic between the hours but still has the error as interpolated between the hours.

For the solar uncertainty case:

- Error generated first with hourly interval
- Deterministic forecast generated with hourly interval
- Uncertain forecast generated by adding error to deterministic
- Uncertain forecast returned to user by interpolation of uncertain forecast with requested interval
- The result at smaller intervals: a forecast that looks like it was generated with hourly interval, but linearly interpolated at the smaller interval.

The reason I still suggest generating solar uncertain forecast at hourly and then interpolate is because there is post-processing done on the result, namely a 3-hour averaging filter, which the developers say in the paper are needed to dampen unrealistic fluctuations in the forecast.  And so, I'm not sure how that should translate to a smaller interval.

HWalnum:
@dhblum Sounds reasonable. I guess the procedure for solar radiation is what mimics a real weather forecast (hourly) best. We could consider doing it the same way for outdoor temperature?

dhblum:
Ok, so resample with linear interpolation what is in-effect an hourly forecast on the temperature too?  Makes sense to me, then at least they are both consistent.

HWalnum:
Yes, I think that is the most consistent solution.

dhblum:
Ok @HWalnum.  The latest commit implements what was suggested.  The uncertain forecast is generated based on an underlying deterministic forecast of hourly interval and error generation of hourly interval, and then interpolated linearly to the interval requested by the user.  Note, the data for the two uncertain variables is still only updated on the hour, although the "time" trajectory returned with the forecast will represent the current time with interval applied through the horizon (as is done currently). With all of this, the return/functionality of all other uncertain variables therefore remains unchanged (with or without an uncertain variable in the same request).

Feel free to test it out.

In the meantime, I'm adding reference results for unit tests.

dhblum:
@HWalnum Thanks so much for your quick reply.  And why it's good to have reviews.

What you say is a bug was actually my intended performance, although perhaps again is flawed. My intention was that indeed the data would not change until the next hour is passed, except for the reported time trajectory (to be consistent with other deterministic forecast variables).  I had thought this makes sense for the values "on the hours", in that they are valid for the duration of the hour until updated again.  This behavior might be suitable in the context where we'd give back constant values over each hour, but not linearly interpolated values between the hours.  In this case, it looks like those values are getting shifted forward.  

Basically, what you're saying, is that the value given at t=900 for the forecast obtained at t=0 should be the same as the value given at t=900 for the forecast obtained at t=900.

HWalnum:
@dhblum I think your idea is valid, if it is meant to be equal to a constant (forward fill) interpolation, but the implementation is wrong, as the whole forecast is shifted. The other issue with this approach is that it will not be consistent with the linear interpolation of the other parameters. 

My idea (as you correctly state) is that we do a linear interpolation, that is consistent with the other variables, so that the user does not have to. I think this makes it much easier for users to use this feature without reprogramming their forecast handling.

dhblum:
@HWalnum I updated the implementation.  Ok to check my work again?

HWalnum:
@dhblum. To me the implementation now looks correct (according to my suggestion). My simulation results also look coherent with the paper.

dhblum:
Thanks a lot @HWalnum.  I am going to update unit test reference results as needed with the updated implementation and then I think this is ready to merge.  Will see if anything else comes up in the meantime.  Do you mind "approving" in the github review interface, unless you have other comments?

dhblum:
Posting similar charts to above with the updated implementation, for reference:

Figure_1

Figure_2

Issue #582, opened 2023-10-18
Reasoning the result of /examples/python/testcase1.py


kefeimo:
## Description
When running the example https://github.com/ibpsa/project1-boptest/blob/master/examples/python/testcase1.py and plot the result, it gives the following plot.

![image](https://github.com/ibpsa/project1-boptest/assets/28743873/b7a3d3d2-f218-4d17-82e0-da5d49e2ffb1)

From GET/measurements
```
{"message":"Queried the measurements successfully.","payload":{"CO2RooAir_y":{"Description":"Zone air CO2 concentration","Maximum":null,"Minimum":null,"Unit":"ppm"},"PHea_y":{"Description":"Heater power","Maximum":null,"Minimum":null,"Unit":"W"},"TRooAir_y":{"Description":"Zone air temperature","Maximum":null,"Minimum":null,"Unit":"K"}},"status":200}
```

from @rlutes "When the PID signal (oveAct_u) goes negative the heating power (PHea_y) should be zero but it is not.  It appears there is something in the model taking the absolute value of the PID signal controlling the heater inside the model."

## Expected outcome
Would like to understand the simulation result and confirm it is behaving as expected.

dhblum:
The performance is as expected, but the documentation of the model and example controller could be improved:

- The HVAC device controlled is actually better characterized as a "heater/cooler" rather than just "heater" since it can provide heating and cooling.
- The Power use of the device is calculated within the model simply as the absolute value of the heating or cooling thermal power delivered divided by an efficiency.
- The "PID" controller implemented in python is really just a "P" controller, as it only implements proportional control.  

I suggest to address these confusions by appropriately fixing naming conventions.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/757.

PR #706, opened 2024-11-15
Issue582 update testcase1 and testcase 3 documentation

Jaap-Neven:
Written basic documentation of the layout of testcase 1 and testcase 3. Included resources to produce a pdf from latex and the most recent pdf version as well.

Did not refactor the names of the basic p-controllers in the example [controllers directory](https://github.com/ibpsa/project1-boptest/tree/de089cb310953a989c32f0c19f66317863a563c3/examples/python/controllers) and corresponding references yet. Should this be done or unnecessary?


dhblum:
This is for https://github.com/ibpsa/project1-boptest/issues/582.

@Jaap-Neven Thanks for the PR!  A couple comments and requests:

1. Instead of latex, pdf, and all the build files, can you provide the documentation of the test cases similarly to the other test cases and as defined in the design spec (see [here](https://ibpsa.github.io/project1-boptest/docs-design/testcasedev.html#documentation))?  That is, add the documentation as an html annotation in the Modelica model (for example see for the bestest_air case [here](https://github.com/ibpsa/project1-boptest/tree/master/testcases/bestest_air/models/BESTESTAir/TestCases/TestCase_Ideal.mo#L29)) and also copy to a ``/testcases/<test_case_name>/doc/index.html``, with any figures in a ``/testcases/<test_case_name>/doc/images/`` directory.  I suggest looking at the other test cases for templates (again see for example the bestest_air case [here](https://github.com/ibpsa/project1-boptest/tree/master/testcases/bestest_air/doc)).  This will align the docs with how things are done in other test cases, and also avoid needing to build, add, and use pdfs in git.
2. One of the things mentioned in #582 was the heater should be considered as a heater/cooler.  Therefore, can you change the measurement variable ``PHea_y`` to ``PHeaCoo_y``?  That will require a change in the Modelica model overwrite block, cascade into needing to update unit test results with the new variable, reflected in documentation, and maybe updates to the scripts in ``/examples``.

dhblum (changes requested):
See previous comment for review requests.

Jaap-Neven:
Hi @dhblum, thanks for the feedback! I'll implement the requested changes at my earliest convenience.

Jaap-Neven:
New FMUs are compiled and all is ready for review

dhblum:
Thanks @Jaap-Neven.  I will merge this to a develop branch and take from here to merge to master.  I'll let you know if anything else is needed from you.

Issue #591, opened 2023-12-04
Include and extract semantic information from BOPTEST testcases


anandkp92:
This issue is to track the development of semantic model inclusion and extraction from BOPTEST testcases. 
The efforts starts with an implementation in the ```multizone_office_simple_air``` testcase. 
- [ ] adding annotations containing semantic information to Modelica model
- [ ] validate information added
- [ ] using modelica-json to extract the semantic model from them
- [ ] validate the semantic model that has been extracted
- [ ] include this workflow into the BOPTEST API

anandkp92:
The development branch to handle this issue is another fork: https://github.com/anandkp92/project1-boptest/tree/issue591_semantics

dhblum:
Thank you @anandkp92.  This is great.  A general comment, the last item about integration into BOPTEST API should focus more on the user's experience of accessing the semantic model/information, whereas the first four items relate to the workflow of a test case developer to include a semantic model in the Modelica and extract for inclusion in the test case.  Just delineating a bit the target audiences/user cases.

dhblum:
@EttoreZ FYI.

PR #709, opened 2024-11-25
Fixing syntax errors in multizone simple office semantic model.

lazlop:
There are some syntax errors in the semantic model it would be good to fix while creating it.  I fixed all the issues, so we can see examples of what needs to be fixed in the track changes. 

Syntax corrections needed are roughly:
- No quotes in node URIs (things after the colon)
- we can replace things in the literal namespace with normal literals now! (literal:text to "text")
- Make sure no periods are in the wrong places.
- Can't use '/' character in URIs. Use qudt for units which has convention X-PER-Y or put in quotes.
- Period only at the end of all triples, if there will be another predicate and object than have to use semicolon.
- Make sure angle brackets don't get through like ref:name literal:<cdl_instance_name>_y
    - Only happened two places

- external references are generally for where a data source is external to the graph. References for zones and equipment  aren't really external, we should discuss if we want to include them.

- has external reference for equipment/zones is incorrect, not actually referencing the external reference but referencing itself. 
```
           e.g. "bldg:flo_Cor a brick:Zone;
          ref:hasExternalReference bldg:flo_Cor."
```
- should flo_Eas be just Eas (and same for other zones)?


EttoreZ:
@lazlop as per conversation during the meeting I updated Issue591 with the latest master version, so you can clone it on your fork and make updates there. Let me know if you have any questions.

dhblum:
Referencing #591.

Issue #712, opened 2024-11-25
Deploy BOPTEST v0.7.0 to Web-Service


dhblum:
This issue is to track deployment of BOPTEST v0.7.0 to the publicly available web-service at https://api.boptest.net.

dhblum:
@kbenne Can you please update/close this issue when completed?

dhblum:
@tijcolem I've tagged a v0.7.1 release via #725.  It looks like the [web](https://github.com/orgs/ibpsa/packages/container/package/boptest-web) and [worker](https://github.com/orgs/ibpsa/packages/container/package/boptest-worker) images have been published as packages with the v0.7.1 tag ok.  Please let me know if this is sufficient for you to continue deployment of the new version.  If so, please continue, and let me know when complete or any other issues.

tijcolem:
@dhblum That's great! I'll post here when the images are pushed to production.

tijcolem:
@dhblum  This is now pushed to prod.  https://api.boptest.net/version  I think we can close this ticket.

dhblum:
Thank you @tijcolem!  One thing though before closing.  When I ping https://api.boptest.net/testcases I get

```
[{"testcaseid":"bestest_air"},{"testcaseid":"bestest_hydronic"},{"testcaseid":"bestest_hydronic_heat_pump"},{"testcaseid":"multizone_office_simple_air"},{"testcaseid":"multizone_residential_hydronic"},{"testcaseid":"singlezone_commercial_hydronic"},{"testcaseid":"testcase1"},{"testcaseid":"testcase2"},{"testcaseid":"testcase3"},{"testcaseid":"twozone_apartment_hydronic"}]
```
which does not include a test case that we newly released in v0.7.0 called ``multizone_office_simple_hydronic``.  I think the ``provision`` service of your production deployment needs to be updated somehow to add this new test case?

tijcolem:
@dhblum  That's true, the update doesn't push new test cases as those are pushed by the provisioner. We can add some notes  in the updates to accommodate this though.

tijcolem:
@dhblum  Yes, let me look over the test case deployment. I think it's just an api push to upload that test case. You can keep this ticket open until I push that.

tijcolem:
@dhblum I uploaded the new test case multizone_office_simple_hydronic and I see it's available now when querying https://api.boptest.net/testcases.  

Did any of the existing test cases change? Should I re-upload all test cases and overwrite?

[...]

PR #722, opened 2025-01-10
Update GitHub Actions to push docker images to Github packages

tijcolem:
The PR adds a Github Action to build, test and publish docker images to GitHub Projects.  Previously, all the logic to test and build the images for the boptest service lived under https://github.com/NREL/boptest-service but has since been merged into this repo. In order to avoid confusion, I suggest that repo can be archived after this is merged. 

In order for the GitHub Action to work under this repo, a GitHub Token and a GitHub Secret must be created.  You must have admin access to perform these steps. 

Steps: 

**1. Generate a token to push images to the GitHub Packages**

-  On [GitHub](https://github.com/) navigate to: 
     User Profile -> Settings -> Developer Settings -> Personal Access Tokens -> Tokens (classic)
- Generate new token (classic) with the following scope: 
    write:packages
Screenshot 2025-01-10 at 12 22 51 PM
- Set an expiration date if desired. 
- Copy the token for the next step

**2)  Create A GitHub Secret.** 
-   [project1-boptest](https://github.com/ibpsa/project1-boptest) -> Settings -> Secrets and Variables -> Actions -> New Repository Secret. 
- Name it "GH_REGISTRY" 

Once those are done, images can then be published to the GitHub Packages for this organization. They will be posted here. https://github.com/orgs/ibpsa/packages

The GitHub Action script will only publish docker images for branches named "develop" and "experimental" and tags (e.g. v0.7.0) for official image releases. The tagged images are used for the production service. All other images are for development.


.github/workflows/build_test_containers.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Run tests
+        run: |
+           cd service
+           docker compose run --no-deps test
+
+      - name: Dump docker logs on failure
+        if: failure()
+        uses: jwalton/gh-docker-logs@v1
+
+#-------------------------  Push to GitHub registry (disabled) -----------------------------
```
dhblum, line 65: This line says "disabled" but the code below is not commented out.  Double check if it should be disabled or adjust comment.

service/ci/publish_to_github.sh
```diff
@@ -1,17 +1,21 @@
 # load .env defines in root of repo
 export $(egrep -v '^#' .env | xargs)
-export GITHUB_WEB_REGISTRY_URI=ghcr.io/NREL/boptest-web
-export GITHUB_WORKER_REGISTRY_URI=ghcr.io/NREL/boptest-worker
+export WEB_REGISTRY_URI=boptest_service-web
+export WORKER_REGISTRY_URI=boptest_service-worker
+export GITHUB_ORG=tijcolem
```
dhblum, line 5: Double checking, should GITHUB_ORG be tijcolem or ibpsa, given step 2 listed in this [comment](https://github.com/ibpsa/project1-boptest/pull/722#issue-2781114341)?

service/ci/publish_to_docker.sh
```diff
@@ -2,8 +2,9 @@
 export $(egrep -v '^#' .env | xargs)
 export WEB_REGISTRY_URI=boptest_service-web
 export WORKER_REGISTRY_URI=boptest_service-worker
-export DOCKER_HUB_WEB_REGISTRY_URI=nrel/boptest-web
-export DOCKER_HUB_WORKER_REGISTRY_URI=nrel/boptest-worker
+export DOCKER_ORG=nrel
```
dhblum, line 5: This is just a question.  Ok, so the actual docker images are hosted under NREL's docker hub account?

dhblum (changes requested):
Thanks @tijcolem for the PR!  Generally looks fine to me, except a few questions and double checks below.

Also, regarding your note "The GitHub Action script will only publish docker images for branches named "develop" and "experimental" and tags (e.g. v0.7.0) for official image releases. The tagged images are used for the production service. All other images are for development." - So any commit to the online ibpsa repo for a branch named "develop" or "experimental" will trigger a new image build, under those tags?

I will work through creating the Token and Secret you outline in Steps 1) and 2).

.github/workflows/build_test_containers.yml
tijcolem: @dhblum  To answer you question, yes, the GitHub Action runner currently will only publish images when there is a push to either develop, experimental and tags. However, the CI will run on any commit for all branches and run the test cases, it just won't publish the images.  

Also, I removed that comment line that said "disabled"

service/ci/publish_to_github.sh
tijcolem: Thanks for catching this!  That was from my testing on my fork. I changed to ibpsa.

service/ci/publish_to_docker.sh
tijcolem: Indeed, but we are now disabling the docker registry and using Github packages for hosting images.

dhblum:
Thanks for all the replies @tijcolem. For step 2) in your original comment, is the token the content that needs to go in the "Secret" dialogue box when creating the New Repository Secret?

dhblum:
Ok @tijcolem I set up the token and secret, and then I restarted the action job.  It seems to be failing as below.  Any suggestions?

Screen Shot 2025-01-14 at 12 16 00 PM

Also just noting generally this PR is for https://github.com/ibpsa/project1-boptest/issues/712.

tijcolem:
Hmm, my suspicion is it could be failing for two reasons. 

1 ) This PR is a from a forked repo and github might have security to prevent this. 
2 ) The ${{ github.actor }} might be using my github user name to try and login to your repository. According to this [doc](github.actor) this variable is "The username of the user that triggered the initial workflow run."    So we might need to change this to use another secret variable.  ${{ secrets.GITHUB_USER }} where that would be your GitHub userid. Not that this is a technically a secret since your GitHub ID is public but I feel good idea to not explicitly use it.  

Currently using: 

```
registry: ghcr.io
username: ${{ github.actor }}
password: ${{ secrets.GH_REGISTRY }}
```
Change to: 
```
registry: ghcr.io
username: ${{ secrets.GH_USERNAME }}
password: ${{ secrets.GH_REGISTRY }}
```

dhblum:
Ok thanks.  Can you push that change to your fork? I'll create the new secret with my username.

tijcolem:
Yep, just pushed the change.

tijcolem:
Hmm, still failing.  Looks like from these [docs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions) secrets (except Github_token) are not passed to forked PRs so this job will fail. 

We change change the PR to merge to a different branch if you want, then merge, and we can test it on that branch not having a forked repo.

dhblum:
Yes, that was going to be my suggestion.

tijcolem:
What branch do you want me to change it to?

dhblum:
Please refer to new PR https://github.com/ibpsa/project1-boptest/pull/723.

PR #723, opened 2025-01-14
Issue712 service deployment

dhblum:
This PR is for #712 and in place of PR https://github.com/ibpsa/project1-boptest/pull/722 for further testing.

PR #724, opened 2025-01-14
Issue712 service deployment 2

dhblum:
For #712 in place of https://github.com/ibpsa/project1-boptest/pull/723.

### DIFF
--- a/.github/workflows/build_test_containers.yml
+++ b/.github/workflows/build_test_containers.yml
@@ -1,0 +1,93 @@
+name: Build, test and publish containers
+
+on:
+  pull_request:
+    branches:
+      -  '*'
+  push:
+    branches:
+      -  '*'
+    tags:
+      - '*'
+
+jobs:
+
+  simulation-tests:
+    name: Build containers and run boptest simulation tests
+    runs-on: ubuntu-22.04
+    timeout-minutes: 240
+    env:
+      COMPOSE_PROJECT_NAME: boptest_service
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.9"
+
+      - name: Build and run stack
+        run: |
+          cd service
+          docker compose build
+          docker compose up -d web worker
+
+      - name: dump docker logs
+        uses: jwalton/gh-docker-logs@v1
+
+      - name: Upload test cases to minio
+        run: |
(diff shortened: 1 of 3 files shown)


dhblum:
@tijcolem Sorry for the various PRs, but I opened this new one to test publishing of the images using a new branch "develop."  It looks like upon merging, the new images were published ok?  See https://github.com/orgs/ibpsa/packages and https://github.com/ibpsa/project1-boptest/actions/runs/12776057731/job/35613840779.

Now, if those look good, I'm realizing from what you said [here](https://github.com/ibpsa/project1-boptest/pull/723#issuecomment-2590977726) "we're good to tag", I wanted to point out that v0.7.0 has already been tagged when we released it in November, and there's been development on the master since that I don't want included in the deployed images if they are to be with reference to v0.7.0.  I'm thinking my options are to either create a new interim tag 0.7.1 and we use those images, or cherry pick your commits that implement the github action to a new branch (maybe "experimental") off of the 0.7.0 tag commit.  But then, I suppose the referenced branch on the image won't be correct.  Do you agree that a new tag 0.7.1 is the best approach here, or other thoughts?

tijcolem:
@dhblum  I'm not seeing the packages on https://github.com/orgs/ibpsa/packages It's possible you need to change the visibility to for those. 

Since you already tagged 0.7.0, a 0.7.1 makes sense in this case.

dhblum:
Got it on the 0.7.1, thanks.

Ok, the packages are published by the organization, and I'm not an admin of the organization (just this repo).  It says the option of making the packages public has been disabled by the organization.  I need to get in touch with an administrator to change permissions.  Alternatively, could I invite you to join the ibpsa organization and then add your username to have access to the packages?  There seems to be an option for that:

Screen Shot 2025-01-14 at 4 33 36 PM

tijcolem:
Cool, I think we'll need to ask to change it to public. It's the k8s manager that will be requesting these images and that lives within AWS. If your org doesn't want to make them public, we can change the k8s to include a user/pass/token to access these images. These were historically made public in the docker hub NREL registry.

dhblum:
Ok thanks.  Let me check in with the organization admin.

dhblum:
@tijcolem The administrator changed permissions to allow public packages and I've made both of the packages, web and worker, public. Please take a look again and let me know if you can access them and test their deployment. Let me know anything else to do, otherwise I will prepare a 0.7.1 release to generate tagged images.

Issue #726, opened 2025-01-18
Misaligned occupancy profile


icupeiro:
The occupancy profile in the testcase 'multizone_office_simple_hydronic' is misaligned and therefore wrong

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/761.

PR #727, opened 2025-01-18
Corrected occupancy profile

icupeiro:
This tackles #726


dhblum:
Thanks @icupeiro for reporting and the PR.  Will review this week when I get a chance.

icupeiro:
More info about this:

This is the plot of the lower setpoint (blue) + occupancy (red), which is completely misaligned

image

Corrected profile is shown here

image

dhblum (changes requested):
@icupeiro I double checked this.  It seems alignment between the occupancy count and zone temperature set point is ok within the FMU simulation.  And, your correction to ``Resources/occupancy.csv`` indeed seems to fix the alignment with what's simulated in the FMU from a time perspective.  However, through further checking, I discovered that it seems:

1. The ``Resources/occupancy.csv`` does not match the decrease of occupancy number in the summer, as seen in the FMU simulation.
2. The ``Resources/internalGains.csv`` does not match the internal gain convective and radiant heat flows for each zone as the FMU simulates (what I believe are variables ``mod.occupancy.heatPortCon[1].Q_flow`` and ``mod.occupancy.heatPortRad[1].Q_flow`` in the wrapped.fmu, plotted below)

See plot below (Note that I only plotted one FMU simulation zone since they are both the same in the simulation.  I suppose I could have done that for the Resources variables too, but plotted both zones to double check).  Can you please confirm if there were updates to the model's occupancy simulation and internal gains that may not have been reflected in the csv's in the Resources directory?  Would you be able to further correct ``Resources/occupancy.csv`` and ``Resources/internalGains.csv`` as appropriate?

Figure_1

dhblum:
@icupeiro Just wanted to check in on this.  It looks like the .csv files in the resources directory haven't been edited in a long time.  I'm wondering, were there updates to the Modelica model that just weren't reflected in the csvs?  Let me know your bandwidth, availability or not, or thoughts on checking in on this over the coming weeks.

icupeiro:
Hi @dhblum 

I pushed what I believe to be the corrected occupancy profile
Regarding point 2, this is normal. The internal gains are a combination of the internal gains in the occupancy model (which represent the appliances) + lighting and occupancy gains, which are embedded in the IDEAS zone model

dhblum:
@icupeiro Thanks.  Confirming things look good with your updated occupancy profile:

Figure_1

dhblum (approved):
Will merge to staging branch and take from there.

Issue #733, opened 2025-02-04
Misaligned occupancy singlezone_commercial_hydronic


SiAndRo2002:
In the singlezone commercial hydronic test case, the occupancy seems to be misaligned to the temperature set points in the typical heating scenario (occupancy during weekend, no occupancy during week days). However, this issue doesn't show in the peak heating scenario where occupancy and temperature set points are in line.

Here is a plot that shows the upper and lower temperature profile as well as the occupancy in the typical heating scenario:
![Image](https://github.com/user-attachments/assets/82d02f67-612a-472c-9fb0-7c5a0d1107b1)

For comparison, here is a plot for the peak heating scenario:
![Image](https://github.com/user-attachments/assets/c0033cdb-369c-44a8-b923-ed14b04461aa)

dhblum:
@HWalnum Is this what you fixed in https://github.com/ibpsa/project1-boptest/pull/703?

HWalnum:
Yes, I think so. I fixed an issue with some missing days in the data. 

However, be aware that this is real occupancy data, so there might be days with low occupancy that are not picked up by the HVAC control (due to holidays).

SiAndRo2002:
Alright, thank you! Then I'll wait for the changes to be approved

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/734.

Issue #702, opened 2024-11-12
Update single zone commercial hydronic with Adrenaline updates


EttoreZ:
This issues is to update test case `single_zone_commercial` with updates coming from Adrenaline

dhblum:
Can this be coordinated with other issues related to this test case?  
- https://github.com/ibpsa/project1-boptest/issues/635 
- https://github.com/ibpsa/project1-boptest/issues/432.

EttoreZ:
@dhblum definitely. I mentioned those two issues in the PR as a reminder.

Issue #635, opened 2024-03-14
singlezone_commercial_hydronic: Unbalanced flow rates supply and exhaust fan


mariuseb:
For the testcase "singlezone_commercial_hydronic": the supply and exhaust fans have unbalanced flow rates, despite the presence of a balancing connection in the infiltration model. A plot of 7 days of operation (first 7 days of the year) that documents the issue is attached.
![fan_issue](https://github.com/ibpsa/project1-boptest/assets/64400580/e38417a4-9a32-4075-9742-de841d38b33e)

EttoreZ:
@mariuseb Thank you Marius. I will take a look.

mariuseb:
@EttoreZ to follow up on this, it appears this issue is resolved by @HWalnum's PR (screenshot of plot attached).
![singlezone_commercial _hydronic_ventilation](https://github.com/user-attachments/assets/c6b70d6d-ea04-4b16-95d8-6c4af41d63a1)

EttoreZ:
@mariuseb thank you for checking! I guess then this issue can be closed once changes from #703 are merged.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/734

Issue #432, opened 2022-06-02
Check mass flows/pressure drops in singlezone_commercial_hydronic


icupeiro:
I think that the mass flow rates in the `singlezone_commercial_hydronic` testcase should be revised. 

We have unsuccessfully tried to do an identification in this model, and the main issue we saw is that the 3-way valve positions (even in the peak period) are always near to be closed (around 1% opening) with the baseline controller. This makes the identification process highly difficult due to the lack of excitation.

Having done a deep dive in the model, I observed that there is only one speed controlled pump, but the pressure drops of the circuit are in the order of a few thousands of Pascals, which seem a bit unrealistic to me looking at the performance curves of the pump model: the resulting mass flow rates in fully open conditions are very very large and represent an idealistic case more than a real case. Therefore, the valve has to work almost closed to produce a large pressure drop and adapt the flow to the set-point.

I would suggest to increase the fixed pressure drops in the circuit to avoid this behavior. An idea is to adapt these to have a resulting mass flow rate at open conditions m_nom = Q_nom/c_p/dT_nom, based on the DH heat exchanger.

dhblum:
Another issue @haraldwalnum reported is that the pump operation may be inconsistent with valve operation.  He reports that currently the main pump is running at constant speed, but the valves operate based on a variable flow system  (fixed pressure difference), so that when there is no heat demand, the pump runs full power against a closed system. @Mathadon offered to potentially take a closer look.

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/734.

PR #734, opened 2025-02-12
Issue702 single zone commercial adrenalin update

dhblum:
This is for #702, following from PR #703 to finalize.  This is also for #635, #432, and #733.


dhblum:
@HWalnum This PR is almost finished.  I didn't see the updates you made to the`` releasenotes.md`` previously but I added them and also marked  the changes in the KPI results for all the scenarios, which I think is good to track if they change significantly version to version.  Can you look at the KPI changes I marked and double check they are what you'd expect or experienced based on your changes?  https://github.com/ibpsa/project1-boptest/pull/734/files#diff-bf0fb780cd5da59cf3fcbb2aeac981b23584b3089c661658f30b386ff4440229.

HWalnum:
Hi @dhblum. For the peak_heat scenario looks reasonable. The changes in the typical_heat scenario seems a bit drastic. I think I might have not updated the days after switching back to the original weather data. However, I tried to have a look with the find_days algorithm. Looking at the result I got from that, I wonder if the median of the daily max is reasonable for commercial buildings with much lower consumption in weekends. This makes the whole algorithm a bit random. 

In the Adrenalin competition we selected the days based more on visual inspection, because the algorithm ending putting one of the scenarios during a holiday, which gave different occupancy profiles.

dhblum:
@HWalnum Hm ok on the day selection algorithm, thanks for reporting.  It certainly isn't as bullet-proof as we thought in the beginning, various issues have come up over time.  Perhaps there can be another option in the script to id and remove weekends.  At least it's documental in terms of the options chosen to produce the days.  Or, we just say that the given days are the time period for a given scenario because the test case developer thought it was reasonable for intent of the scenario.

HWalnum:
I definetly see the benefit of having a documentable procedure, but it is challenging to make it fit all cases. For typical heat day, it could be an idea to look at weekly averages. However, it will not remove the risk of including special holdiays etc. However, the last is mainly a problem for cases like this, where real measured occupancy is used. 

For this particular case, I would suggest to go for 308 as typical_heat_day. This is based on median of weekly averages, and it should not coincide with any holidays.

dhblum:
Ok thanks @HWalnum.  Then 308 sounds good and I will plan to prep the PR using that.

dhblum:
Hey @HWalnum I've updated the days.json with the new typical heat period, updated unit test ref results, and updated the KPI changes.  Take a look at https://github.com/ibpsa/project1-boptest/pull/734/commits/456d885b906c17e64f55234f8244f55b9077e3f3 and see if the new change values are reasonable to you.  I think they seem ok.  Otherwise, just waiting for unit tests to pass.

HWalnum:
@dhblum Looks good to me :)

dhblum:
Great thanks.  Once unit tests pass, I'll push the new test case FMU and finalize for merging.

Thank you so much for your help and effort on this.

HWalnum:
Shit. Sorry about that. 
The changes should have been done in DistrictHeating2. It was only adding the two read blocks (+ modifying the description in the overwrite block). So it should not require changes to the documentation. 
I should have deleted the old models from the repository. Maybe we should do that now.
Do you want me to do the changes, or do you want to do it yourself, in you PR?

dhblum:
No worries.  Can you make a PR to branch ``issue702_single_zone_commercial_adrenalin_update`` with the changes to the Modelica models, including deleting old models to prevent confusion in the future?  

I already committed changes to the docs that used the (now incorrect) recompiled FMU (I noticed something that made me want to do that).  It's ok, after your new PR, I'll re-update things, and thereby double checking it's all consistent.

dhblum:
Alright @HWalnum thanks.  I merged PR https://github.com/ibpsa/project1-boptest/pull/737 and updated docs, bacnet.ttl, and unit test results.  Just waitting for unit tests to pass and then will again update wrapped FMU and mo.  Then we should be ready to merge :)

Thanks again.

Issue #748, opened 2025-04-14
podman support for rootless environments


acedrew:
Trying to get the quick start working with podman, initial issues:
- Privileged default ports - this is a simple fix, just requires making sure all the ports are plumbed through to use the env variables
- Docker vs Podman env handling - this was a bit of a headache, the podman requires interpolation to use .env vars, not just declaring them in the instance, took me a while to figure out
- S3 - Probably related to the environment variable issue, but web is trying to initialize S3 and gets an invalid access key id, despite being pointed at the minio instance. Still not sure where this is happening, and attempts to log it haven't worked out so far.

dhblum:
Thanks @acedrew.  @kbenne do you have any insight or experience on deployment with podman?

- The PR 749 @acedrew drafted suggests changing the web port to listen on :8000 rather than :80.  I've heard from a couple other folks that :80 tends to be restricted or privileged in some way.  Was there a particular reason for choosing :80 to start with?  Could moving to something like :8000 be beneficial?
- Seems like the S3 issue is the biggest one here.

PR #749, opened 2025-04-14
WIP: Changes to allow manipulation of the server port via the .env file to…

acedrew:
… allow use of unprivileged ports in rootless environments
Addresses #748


acedrew:
This is running on my end! Testing should probably load the port from the same default config in the repo that is building the service

Issue #764, opened 2025-05-19
Bacnet interface update


EttoreZ:
Currently the bacnet interface automatically runs the test case simulation in real time, advancing 5 seconds of simulation every 5 seconds. This issue is to add options to deploy and advance time:

1. Every X seconds of simulation time every Y seconds of real time where X and Y are parameters
2. Upon trigger, which is how the native BOPTEST API works.

This issue will enhance BOPTEST Bacnet interface with the following features:

- For 1., add command line parameter to [BopTestProxy.py](https://github.com/ibpsa/project1-boptest/blob/master/bacnet/BopTestProxy.py) to allow user to set simulation time step and Bacnet appinterval time.
- For 2., add feature in [BopTestProxy.py](https://github.com/ibpsa/project1-boptest/blob/master/bacnet/BopTestProxy.py) to allow user to manually advance simulation.
- Add simulation time as an available bacnet point to read

dhblum:
@EttoreZ Thanks for opening.  At least, I suggest splitting 3. into it's own issue separate from 1 and 2.  I'd like to separate the implementation of time management from activate/deactivate so as not to have to compound things too much at once and have one block the completion of the other.

EttoreZ:
> [@EttoreZ](https://github.com/EttoreZ) Thanks for opening. At least, I suggest splitting 3. into it's own issue separate from 1 and 2. I'd like to separate the implementation of time management from activate/deactivate so as not to have to compound things too much at once and have one block the completion of the other.

@dhblum I thought we could have 1 issue to coordinate BACnet interface updates and multiple PRs for the different tasks, but I can to remove task 3 from this issue.

dhblum:
Just noting that deactivate/activate is already handled in the BACnet proxy since all activation signals are first set to 0 and then only set to 1 if valid value is seen from BACnet.  See [Lines 158-184](https://github.com/ibpsa/project1-boptest/blob/master/bacnet/BopTestProxy.py#L158).

dhblum:
Closed by https://github.com/ibpsa/project1-boptest/pull/765.

PR #765, opened 2025-05-22
Issue764 bacnet interface update

EttoreZ:
This PR is to address issues #764.
The main changes to the script `bacnet/BopTestProxy.py` are:

- Introducing optional command line arguments 'app_interval' and 'simulation_step'. These allow to change the frequency at which the simulation is advanced 'app_interval' and by how long 'simulation_step'. If  'simulation_step' > 'app_interval' the simulation will run faster than real time
- changed from using function decorator '@recurring_function(APPINTERVAL)' to instantiating the `RecurringTask` module within the script itself. This allows to initialize 'APPINTERVAL' with the command line argument, and it allows to handle the 'oncommand' simulation.
- If 'app_interval' is set to 'oncommand' value the simulation adds the variable 'advance' to the available inputs in BACnet object. If 'advance(t)' > ' 'advance(t-1)' the simulation will advance.


dhblum (changes requested):
Thank you @EttoreZ for the implementation.  Generally, it looks good.  I made some edits though and pushed them, including merging latest master.  We're on the right track I think, so can you address the following additional development:

- [x] Implement a mechanism to restart the 'advance' counter so that a user doesn't end up needing to send a very very large number after operations for long time periods.
- [x] Implement a new bacnet output point from the proxy that provides the client with the current simulation time.  This will be useful for users now that they have the capability to manipulate time in more complex ways than the original implementation.  At this point I suggest returning a time stamp to be saved for a future issue because it is more involved especially regarding timezones, DST, leap years, etc.
- [ ] Implement new unit tests in the file ``testing/test_bacnet.py`` to test for each configuration of the proxy in terms of time manipulation (default settings, automatic advance with new step, user-defined advance with new step).  I therefore see at least two new tests.  It would be good to also test that time is returned properly to a client.  I forget how familiar you are with the unittest package in Python, but we can chat offline if you need a bit more guidance on how to implement the new tests. 
- [ ] Update the ``bacnet/readme.md`` file with the new time management options available upon proxy deployment, and also about the return of simulation time as a bacnet point.
- [ ] Update the ``releasenotes.md`` with a bullet regarding the updates implemented in this PR.

dhblum (approved):
This PR is ready to merge pending any final thoughts.

EttoreZ:
> @EttoreZ Tests passed on your recent commit and I just made a couple updates to the readme and release notes. Are you finished with changes and is this ready to merge in your eyes?

Ready to merge @dhblum!

PR #803, opened 2025-10-28
Issue732 large results payload

kbenne:


dhblum (changes requested):
Thanks @kbenne.  Generally this looks good, and the threshold is good to allow users to still use /results at high frequency for smaller data packages, and seems to work.  Two questions:

1.  What is the max amount of data recommended or allowable through redis?  Your current threshold is 256 kB to switch to S3.  But, in the existing approach, we have an example (our tutorial) where about 10x more data is being passed through redis successfully already.

2. Can you fix so that if S3 is used, the bytes are not printed to the logger?  See the screen shot below, where the unintelligible block of text extends beyond the limits of the terminal in this case even if I scroll all the way to the top.

Screen Shot 2025-10-28 at 5 29 05 PM

dhblum (approved):
Regarding my comment above:

1. I increased the redis message limit to 1 MB, per this [article](https://aws.amazon.com/compare/the-difference-between-rabbitmq-and-redis/) and through some testing that showed reasonably similar query times using redis and object storage.  From observations, note that the smaller the data package, the more redis outperforms object storage in terms of query speed.  Changed in this commit: https://github.com/ibpsa/project1-boptest/pull/803/commits/ad6bf2c94a9252540a4e3e2acf26d8b042b12901.

2. I was able to suppress the nonsensical and large log print by adjusting the log levels for the boto3 package, in this commit: https://github.com/ibpsa/project1-boptest/pull/803/commits/ad6bf2c94a9252540a4e3e2acf26d8b042b12901.

To Do:
- [ ] Confirm new limit is ok or if should use lower.  Update documentation as needed.
- [x] Update release notes
- [ ] Unit tests pass

dhblum:
Thank you @kbenne!  Then I will update some of the documentation and then I think this is ready to go!

Issue #819, opened 2025-12-23
Add OpenAPI Spec and Docs


dhblum:
This issue is to add an [OpenAPI spec](https://swagger.io/specification/) of the BOPTEST API for controller testing.  Adding this will:

- enable systematic tracking of the API implementation 
- add a machine-readable API spec 
- enable auto-generation of human-readable and interactive API docs 
- enable auto-generation of a Python client that could be used for unit testing of the spec

PR #815, opened 2025-11-18
Add openapi spec and docs

bertrandkerres:
*As discussed on today's (2025-11-18) BOPTEST meeting.*

This PR adds an [OpenAPI specification](https://editor.swagger.io/) to the `web` service.

Screenshot 2025-11-18 at 20 23 49

Benefits:
* Human and machine readable `yaml` file with description of all common endpoints for controller development
* This includes expected input and output format, as long as it is not testcase-specific
* Automatic generation of human readable docs, hosted at `http://<boptest-url>/docs` (e.g. it would be http://api.boptest.net/docs after deployment). This requires 2 added packages, `swagger-ui-express` and `yamljs`.
* The docs are interactive: One can use them to send HTTP requests to the server.
* The file can be used to automatically generate clients (software that uses BOPTEST via REST API), e.g. using [openapi-python-client](https://github.com/openapi-generators/openapi-python-client) or [heyapi](https://heyapi.dev/) (TypeScript)

Limitations:
* The file is currently hardcoded (mostly written by LLMs), so it would need to be updated everytime the BOPTEST API is updated. But that shouldn't happen too often.
* The administrative endpoints are currently not part of the OpenAPI file.

Issue #760, opened 2025-05-03
CORS headers to use BOPTEST from browser


bertrandkerres:
Related to the architecture described in #759 , there's also an issue with accessing BOPTEST from a browser related to [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS). Essentially, when contacting BOPTEST from a web browser, the browser will refuse to fetch from BOPTEST (if the browser doesn't run in unsafe mode).

In order to allow browsers to fetch and display BOPTEST data, the BOPTEST web service would need to:
* Respond to preflight requests (`OPTIONS {endpoint}`, e.g. `OPTIONS /results` with some specific headers) with a correct response
* Add a header entry to the actual request (e.g. `PUT /results`)

I've built two small middleware handlers that solve the problem for the test case interaction endpoints, but I don't know about the administrative endpoints like `/users/:userid` etc that might also require authentication.

In `service/web/server/src/routes/middleware.js`:
```javascript
export function handleCorsPreflight(req, res, next, allowedMethods) {
  if (req.method === 'OPTIONS') {
    const origin = req.header('Origin');
    res.setHeader('Vary', 'Origin');

    if (origin && isValidOrigin(origin)) {
      console.log("Pre-flight allowed: ", origin)

      res.setHeader('Access-Control-Allow-Origin', origin);
      res.setHeader('Access-Control-Allow-Methods', allowedMethods);
      res.setHeader('Access-Control-Allow-Headers', 'Content-Type, Authorization');
      res.setHeader('Access-Control-Max-Age', '86400'); // Cache preflight response for 24 hours
      res.sendStatus(204); // No Content
    } else {
      console.log("Pre-flight rejected: ", origin)
      res.sendStatus(405); // Method Not Allowed
    }
  } else {
    next();
  }
}

export function addCorsOriginHdr(req, res, next) {
  const origin = req.header('Origin');
  if (origin && isValidOrigin(origin)) {
    res.setHeader('Access-Control-Allow-Origin', origin);
    res.setHeader('Vary', 'Origin');
  }
  next()
}

function isValidOrigin(origin) {
  const allowedOriginRegex = /^http:\/\/localhost:\d+$/; // e.g. all http://localhost:port
  return allowedOriginRegex.test(origin)
}
```

In `service/web/server/src/routes/boptestRoutes.js`:
```javascript
// Example for PUT /forecast

/*
 Register OPTIONS pre-flight
[...]

PR #820, opened 2026-01-02
Issue819 open api spec

dhblum:
This is for #819 and continues from PR #815.  Todos:

- [x] Editorial tweaks
- [x] Add KPIResponse schema
- [x] Rearrange/Refine "tags"
- [x] Add unit test with auto-generated Python client
- [x] Edits to time and kpi properties from comments
- [x] Confirm unit test to make sure it catches meaningful errors
- [x] Version change (see also https://github.com/ibpsa/project1-boptest/issues/826)
- [x] Reference /docs on README.md
- [x] Shorten "tags"
- [x] Update release notes and contributors
- [x] Final reviews

FYI @bertrandkerres.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ BOPTEST is also available as a public web-service and can be used by following t
 1) Use the API below to first select a test case to run, and then interact with it using your test controller.  Send API requests to ``https://api.boptest.net/<request>``
 
 ## RESTful HTTP API
-This API is documented using the [OpenAPI Specification](https://swagger.io/specification/) in ``service/web/server/docs/openapi.yaml``.  You can view the full API using Swagger docs by deploying BOPTEST as described above and going to ``http://127.0.0.1:8000/docs`` in a browser.  A summary of the API is provided in the table below.
+This API is documented using the [OpenAPI Specification](https://swagger.io/specification/) in ``service/web/server/docs/openapi.yaml``.  You can view the full API using Swagger docs by deploying BOPTEST as described above and going to ``http://127.0.0.1:8000/docs`` in a browser.  You can also access the specification at ``http://127.0.0.1:8000/openapi.yaml``. A summary of the API is provided in the table below.
 
 API requests that interact with a running test case (those that require a ``testid``) will return a JSON in the form ``{"status":<status_code_int>, "message":<message_str>, "payload":<relevant_return_data>}``. Status codes in ``"status"`` are integers: ``200`` for successful with or without warning, ``400`` for bad input error, or ``500`` for internal error.  Data returned in ``"payload"`` is the data of interest relvant to the specific API request, while the string in ``"message"`` will report any warnings or error messages to help debug encountered problems.
 


testing/makefile
```diff
@@ -309,6 +309,19 @@ test_testcase:
...
 	fi
 
+test_autogen_client:
+# Test automatic generation of python client from OpenAPI Spec
+# Build and deploy testcase image
+	cd .. && docker compose build -q web worker provision
+	cd .. && docker compose up -d --quiet-pull web worker provision
+	python sleep10.py
+# Run openapi-python-client to autogenerate api client
+	openapi-python-client generate --path ../service/web/server/docs/openapi.yaml --output-path ../examples/python/autogenerated_client/output $(ARG)
```
bertrandkerres: You might need to activate the venv to run `openapi-python-client`, if it is not done by some other process (I don't know how Travis CI works)

testing/makefile
```diff
@@ -309,6 +309,19 @@ test_testcase:
...
+# Run the autogenerated client using the example usage.py
+	python ../examples/python/autogenerated_client/usage.py
```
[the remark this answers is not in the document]
dhblum: A few tests and associated terminal output show (see below) that the ``try:finally`` statement doesn't prevent an error from being thrown (and the makefile returning a code of 1 indicating error, which will trigger travis to fail too).  It's because there's no ``except`` statement.  The ``try`` effectively does nothing here without an ``except``.  If I add an ``except`` just to print "an error happened", the ``finally`` statement is triggered afterward, and no error is thrown (and makefile returns 0 indicating no error, and in this case travis would pass).

Can't think of another solution at the moment. But I might be ok leaving as is, since a user could add their own ``except`` clause.  Or, we add one that still actually throws an error, but acknowledge it's not a graceful termination (as the script stops with an error) and a user may want to replace it with some other handling.

Some error tests and terminal outputs
--------------------------------------
Try to access sim_res.payload["timee"] when should be "time"
```
KeyError: 'timee'
make: *** [makefile:319: test_autogen_client] Error 1
```

Try to access sim_res.payload.time[:5] when should be ["time"]
```
AttributeError: 'TimeSeries' object has no attribute 'time'
make: *** [makefile:319: test_autogen_client] Error 1
```

Try to use put_forecast when should be put_forecast_testid
```
ImportError: cannot import name 'put_forecast' from 'output.boptest_service_api_client.api.test_simulation_and_data'
make: *** [makefile:319: test_autogen_client] Error 1
```

service/web/server/docs/openapi.yaml
```diff
@@ -0,0 +1,679 @@
+openapi: 3.0.3
+info:
+  title: BOPTEST Service API
+  description: A REST API for emulating buildings for testing control systems.
+  version: 1.0.0-dev
+servers:
+  - url: http://localhost
```
bertrandkerres: http://localhost:8000 with the new updates

service/web/server/src/index.js
```diff
@@ -31,6 +33,10 @@ app.use(express.urlencoded({ extended: true }))
 app.use('/', boptestRoutes)
 app.use(errorHandler)
 
+// Serving the docs using swagger
```
bertrandkerres: Would be good to also serve the openapi.yaml file itself. This allows people to e.g. autogenerate the client by just pointing to http://api.boptest.net/openapi.yaml as input, they don't need to extract the file from the repo.

bertrandkerres (commented):
I only found small comments in the code, made a PR to fix them directly.

One related, bigger issue not yet discussed is CORS headers. This is something only relevant when interacting with BOPTEST through a browser, so it was not a big issue until now. However, since we now have docs in the browser and would like people to interact with it, it would be good to fix it.
What happens is that when a browser sends a request to another domain, it always adds a HTTP header `"Origin"` with the domain it's coming from. The server should then approve that the browser origin domain is acceptable. If the server doesn't send the approval message, the browser will refuse to process the server response.
You can see the problem if you open the docs on `http://localhost:8000/docs`, and then choose `http://api.boptest.net` as server in the docs dropdown menu, and then try interacting with the server. So now the browser sends `"Origin": "localhost"`, doesn't get a response, and refuses to process the request.

I had made an issue a while back, https://github.com/ibpsa/project1-boptest/issues/760, should we put that into this branch as well? (I already have the code to fix it, since I used it locally)

